PR #11, opened 2020-04-23
Allow for pluggable public dirs and service ui skins

Fryguy:
@simaishi Please review.

Depends on https://github.com/ManageIQ/manageiq/pull/20120

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -1,3 +1,4 @@
+require 'json'
 require 'pathname'
 require 'yaml'
 
@@ -5,10 +6,12 @@ module ManageIQ
   module RPMBuild
     class GenerateCore
       include Helper
-      attr_reader :miq_dir
+
+      attr_reader :miq_dir, :ui_service_dir
 
       def initialize
         @miq_dir = BUILD_DIR.join("manageiq")
+        @ui_service_dir = BUILD_DIR.join("manageiq-ui-service")
       end
 
       def build_file
@@ -21,6 +24,10 @@ module ManageIQ
 
       def release_file
         File.write(miq_dir.join("RELEASE"), RELEASE)
+      end
+
+      def link_plugin_public_dirs
+        symlink_plugin_paths("public", miq_dir.join("public"))
       end
 
       def precompile_assets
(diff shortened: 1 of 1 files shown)


Fryguy:
Note that I think from a development perspective we need some of these things as rake tasks from the individual repos, however for now we can keep it here, and then extract those later.

build_scripts/manageiq_tar_build.rb
```diff
@@ -59,7 +65,9 @@ def precompile_sti_loader
   end
 
   def build_service_ui
-    Dir.chdir(app_dir.join("manageiq-ui-service")) do
+    symlink_plugin_paths("manageiq-ui-service", ui_service_dir)
```
Fryguy, line 68: This, in particular should probably be a rake task as part of manageiq-ui-service itself (but future PR)

simaishi:
This fails on the build machine...

```
Could not load database configuration. No such file - ["config/database.yml"]
```

Fryguy:
oh it must need environment for rails runner...I'll have to write a dedicated rake task.

build_scripts/manageiq_tar_build.rb
```diff
@@ -1,13 +1,15 @@
+require 'fileutils'
 require 'pathname'
 require 'yaml'
 
 class ManageIQTarBuild
-  attr_reader :app_dir, :cfg_dir, :miq_dir, :options
+  attr_reader :app_dir, :cfg_dir, :miq_dir, :ui_service_dir :options
```
chessbyte, line 6: missing a comma?

Fryguy:
Depends on https://github.com/ManageIQ/manageiq/pull/20120

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -23,6 +26,10 @@ def release_file
         File.write(miq_dir.join("RELEASE"), RELEASE)
       end
 
+      def link_plugin_public_dirs
+        symlink_plugin_paths("public", miq_dir.join("public"))
```
Fryguy, line 30: @simaishi I noticed the following line of code, which might cause this to be an issue:

https://github.com/ManageIQ/manageiq-rpm_build/blob/a8d37edafeba9d751c53a6acb72de47a46d71bd2/lib/manageiq/rpm_build/generate_tar_files.rb#L22

Can we discuss this?  I think I understand why, but I'm thinking there might be a better way to do it there, so we don't have to enumerate things individually.

simaishi: https://github.com/ManageIQ/manageiq-rpm_build/pull/18 removes those 2 lines. Working on PR feedback, should be ready for another review soon.

Issue #4, opened 2020-04-21
rpm version/release number update for nightly/tag builds


simaishi:
The rpm is currently using `Release:  1.<build_date>%{?dist}` for the rpm 'release' in N-V-R.

This needs to be changed to:

- include manageiq core repo sha for nightly build (I think...)
- include the tag name for tag build

Fryguy:
I'm ok if the NVR only has the numerical version and we include the SHA and tag name *inside* the build (I think we already do in the BUILD/VERSION/RELEASE files)

simaishi:
@Fryguy took what we've discussed and changed a bit based on the guidelines (https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/)

jansa (nightly) - 10.1.0-0.1.yyyymmddhhmmss
jansa-1-alpha1 - 10.1.0-0.2.alpha1
jansa (nightly) - 10.1.0-0.2.yyyymmddhhmmss
jansa-1-beta1 - 10.1.0-0.3.beta1
jansa-1 - 10.1.0-1
jansa-2 - 10.2.0-1
jansa-2.1 - 10.2.1-1
jansa-2.1 (but with RPM change) - 10.2.1-2

The nightly isn't following 100% as we don't include git sha, but packages will be updated properly with the above.

```
$ rpmdev-vercmp 10.1.0-0.1.20200423000000.el8 10.1.0-0.2.alpha1.el8
10.1.0-0.1.20200423000000.el8 < 10.1.0-0.2.alpha1.el8

$ rpmdev-vercmp 10.1.0-0.2.alpha1.el8 10.1.0-0.2.20200424000000.el8
10.1.0-0.2.alpha1.el8 < 10.1.0-0.2.20200424000000.el8

$ rpmdev-vercmp 10.1.0-0.3.beta1.el8 10.1.0-1.el8
10.1.0-0.3.beta1.el8 < 10.1.0-1.el8
```

Fryguy:
What is the number after the -0, but before the timestamp? Is it necessary?  I'm concerned it will get complicated keeping that value stored somewhere for successive nightly builds.

simaishi:
That's necessary to keep the update work. Without it, digits (timestamp) > pre-release name (alpha, beta, etc.) and you won't be able to update from nightly to pre-releases.

I'm thinking we can have a rake task to increment that digit when we tag, just like how we modify VERSION file for the core repo.

Fryguy:
[...]

Issue #8, opened 2020-04-22
Allow passing build options to docker run


simaishi:
At minimum,`git_tag` needs to be overriden, to fully automate both nightly/tag builds.

PR #16, opened 2020-04-27
Add support for nightly/tag builds and non-interactive build

simaishi:
- Add non-interactive build from `docker run` command

  `docker run <miq-rpm_build image> build [--build-type <type>] [--git-ref <ref>]`

  Running without "build" argument will behave like before: setup build environment and load bash.

- By default, build will run for 'nightly' with branch specified in config/options.yml. This can be overridden by `--build-type` and `--git-ref` options. Those 2 are what need to change between nightly and tag builds for a given branch.

  `bin/build.rb --build-type release --git-ref jansa-1-alpha1`

The RPM format is `<version>-<rpm:release>.<nightly/pre-release info>`
Example:

build type | version | rpm:release | nightly/pre-release | RPM NVR
---------- | ------- | ------------ | ------------------- | ---------------
nightly | 10.1.0 | 0.1 | \<date\> | 10.1.0-0.1.20200427130019
pre-release | 10.1.0 | 0.2 | beta1 |  10.1.0-0.2.beta1
GA | 10.1.0 | 1 | | 10.1.0-1

`version` and `rpm:release` will be updated when repo gets tagged (a follow up, issue #17)

Fixes https://github.com/ManageIQ/manageiq-rpm_build/issues/4 and https://github.com/ManageIQ/manageiq-rpm_build/issues/8


bin/build.rb
```diff
@@ -3,9 +3,19 @@
...
+require 'optimist'
+
+opts = Optimist.options do
+  opt :build_type, "nightly or release", :type => :string, :default => "nightly"
+  opt :git_ref,    "Git ref to use (default: git_ref specified in options.yml)", :type => :string
+end
+
+build_type = opts[:build_type]
+git_ref    = opts[:git_ref]
+Optimist.die "git_ref must be set for release build" if build_type == "release" && git_ref.nil?
```
Fryguy, line 15: If you specify the `git_ref` in the options yml, and then set `--build_type=release` then this will fail.

I think the line above needs to be something like `opts[:git_ref] || config_options[:git_ref]`, or perhaps just drop this check altogether.

bin/build.rb
```diff
@@ -3,9 +3,19 @@
...
 
 require 'manageiq/rpm_build'
+require 'optimist'
+
+opts = Optimist.options do
+  opt :build_type, "nightly or release", :type => :string, :default => "nightly"
+  opt :git_ref,    "Git ref to use (default: git_ref specified in options.yml)", :type => :string
+end
+
+build_type = opts[:build_type]
```
Fryguy, line 13: I think you should add

```ruby
Optimist.die "build type must be either nightly or release" unless %w[nightly release].include?(opts[:build_type])
```

container-assets/user-entrypoint.sh
```diff
@@ -3,7 +3,17 @@
 export GEM_HOME=$(ruby -e 'puts Gem.user_dir')
 export PATH=$GEM_HOME/bin:$HOME/bin:$PATH
 
+if [[ $# -eq 0  ]]; then
```
Fryguy, line 6: Minor, should only be one space after the `0`

lib/manageiq/rpm_build/setup_source_repos.rb
```diff
@@ -5,13 +5,13 @@
 module ManageIQ
   module RPMBuild
     class SetupSourceRepos
-      attr_reader :git_tag, :github_url
+      attr_reader :git_ref, :github_url
 
-      def initialize
+      def initialize(ref)
         puts "\n---> #{self.class.name}::#{__method__}"
         options = YAML.load_file(CONFIG_DIR.join("options.yml"))
```
Fryguy, line 12: Maybe this could be moved to a top-level ConfigOptions class (or use the config gem)...then you can access those options from the bin script.

bin/build.rb
simaishi: That's true... it doesn't make sense to check if default git ref is in options.yml file. I'll drop this.

lib/manageiq/rpm_build/setup_source_repos.rb
Fryguy: nvm, since you are going to drop the check it's not needed for this PR(though maybe that could be a separate refactoring).

Issue #5, opened 2020-04-21
Review package "Requires" for rpms


simaishi:

simaishi:
The initial review was done during moving podified to rpm based. Any further changes can be tracked separately.

PR #18, opened 2020-04-30
Change to build "manageiq" package with multiple subpackages (RPMs)

### LINKED ISSUE #9: Ensure 3 RPMs used in appliance build are from the same build
Currently, 3 RPMs (manageiq, manageiq-gemset and manageiq-appliance) are built independently in Copr and any one of them could fail.

We need to put some kind of mechanism to make sure appliance build will pick up matching RPMs.

The easiest way is to "Require" the exact same version-release. Alternatively, we can have 1 RPM spec file to build all RPMs as sub packages.

simaishi:
As discussed in https://github.com/ManageIQ/manageiq-rpm_build/issues/9, it now builds "manageiq" package with the following subpackages

- manageiq-appliance (top-level package for appliance build)
- manageiq-appliance-tools
- manageiq-core
- manageiq-gemset
- manageiq-pods (top-level package for manageiq-base build)
- manageiq-system
- manageiq-ui (contains public/{assets, packs, ui})

The build will continue creating 3 tarballs, but "manageiq" was renamed to "manageiq-core".

The rpm spec file was getting big, so I created separate files for each subpackage, which will be combined into the main .spec during the build (or by running bin/generate_spec.rb).

This PR moves existing files and package "requires" to new subpackages, so it might not be correct if it wasn't correct before this PR. That will be reviewed as a part of #5.

Note:
- manageiq-pods isn't tested yet. This will be tested after #5 is completed.
- This PR is being tested with kickstart file changes in https://github.com/simaishi/manageiq-appliance-build/tree/rpm_build_kickstart

Fixes #9


bin/generate_spec.rb
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env ruby
+
+require 'pathname'
+
+spec_dir = Pathname.new("../rpm_spec").expand_path(__dir__)
+manageiq_spec = spec_dir.join("manageiq.spec.in").read
+
+Dir.glob("#{spec_dir.join("subpackages/*")}") do |spec|
```
Fryguy, line 8: Minor, but you might want to do `Dir.glob().sort` so that you get consistent output.

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -54,32 +54,21 @@ def seed_ansible_runner
...
-        # Add a product_name-version directory to the top of the files added to the tar.
-        # This is needed by rpm tooling.
-        transform = RUBY_PLATFORM =~ /darwin/ ? "-s " : "--transform s"
-        transform << "',^,#{tar_basename}/,'"
-
-        # Everything from */tmp/* should be excluded, except for tmp/cache/sti_loader.yml
-        system_cmd("tar -C #{miq_dir} #{transform} --exclude-tag='cache/sti_loader.yml' -X #{exclude_file} -hcvzf #{tarball} .")
-        puts "Built tarball at:\n #{File.expand_path(tarball)}"
+        if ENV["NPM_REGISTRY_OVERRIDE"]
+          Dir.chdir(BUILD_DIR.join("manageiq")) { system_cmd("#{SCRIPT_DIR.join("scripts/npm_registry/yarn_registry_cleanup.sh")}") }
```
Fryguy, line 70: `BUILD_DIR.join("manageiq")` is just `miq_dir`

lib/manageiq/rpm_build/generate_tar_files.rb
```diff
@@ -19,50 +13,37 @@ def shell_cmd(cmd)
...
-        Dir.chdir(BUILD_DIR) do
-          transform = "--transform s',\^,#{PRODUCT_NAME}-appliance-#{VERSION}\/,\'"
-          base_dir = BUILD_DIR.join("manageiq-appliance")
-          shell_cmd("tar -C #{base_dir} #{transform} --exclude='.git' -hzcf #{PRODUCT_NAME}-appliance-#{VERSION}.tar.gz .")
 
-          FileUtils.cp(BUILD_DIR.join("#{PRODUCT_NAME}-appliance-#{VERSION}.tar.gz"), RPM_SPEC_DIR.join("#{PRODUCT_NAME}-appliance"))
-        end
+        tar_basename = "#{PRODUCT_NAME}-appliance-#{VERSION}"
+        tar_name  = RPM_SPEC_DIR.join("#{tar_basename}.tar.gz")
+        transform = "--transform s',\^,#{tar_basename}\/,\'"
```
Fryguy, line 32: I code elsewhere I thought we had a check for Mac vs other here?

lib/manageiq/rpm_build/generate_tar_files.rb
```diff
@@ -19,50 +13,37 @@ def shell_cmd(cmd)
...
-        tar_build.tar_prep
-
-        if ENV["NPM_REGISTRY_OVERRIDE"]
-          Dir.chdir(BUILD_DIR.join("manageiq")) { shell_cmd("#{SCRIPT_DIR.join("scripts/npm_registry/yarn_registry_cleanup.sh")}") }
-        end
-
-        tar_build.tar
+        tar_basename = "#{PRODUCT_NAME}-core-#{VERSION}"
+        tar_name = RPM_SPEC_DIR.join("#{tar_basename}.tar.gz")
+        transform = "--transform s',\^,#{tar_basename}\/,\'"
```
Fryguy, line 42: Is it possible to move this into a method so we don't have to duplicate it with the code above?

lib/manageiq/rpm_build/generate_tar_files.rb
simaishi: This was the only place it was checking, so I just removed it... especially now with a container image that can run the build, I thought it's not needed. I never tried running on Mac, so I'm not sure if transform is the only thing that needs the check, but I can add if needed.

Fryguy: oh interesting...if I recall some part of the `--transform` thing was added because on a Mac the tar command works slightly differently.

Fryguy: So, yeah, now that we are container based, we can ignore Mac specifics.

simaishi:
Rebased and added 2 more commits.

PR #22, opened 2020-05-01
Pin to UBI 8.1

Fryguy:
@simaishi Please review.

This is similar to the issue we ran into in https://github.com/ManageIQ/manageiq-pods/pull/474


Fryguy:
Fixes #25

PR #23, opened 2020-05-01
Move bundle into docker build

Fryguy:


simaishi:
There was a reason bundle was called after setting GEM_HOME in user-entrypoint.sh. It was about gems going to non expected location(?), but I don't remember the details. So I'll have to run it and check...

Fryguy:
This is the gems for the manageiq-rpm_build code itself, as opposed to the gems for the build.  However, if there's an overlap, maybe that can cause issues?

I'm also noticing this line which I don't understand, but might be related? - https://github.com/ManageIQ/manageiq-rpm_build/blob/a8d37edafeba9d751c53a6acb72de47a46d71bd2/lib/manageiq/rpm_build/generate_tar_files.rb#L51

simaishi:
> This is the gems for the manageiq-rpm_build code itself, as opposed to the gems for the build. 

Yeah, I know... `rake` is related, I think. It wasn't finding bundler and/or rake , and trying to set it differently caused GEM_HOME for the manageiq-gemset to be overriden.

Fryguy:
closing for now, and we can discuss later.

Issue #2, opened 2020-04-21
Add productization support


simaishi:

Fryguy:
So far, outside of a normal plugin the only things I've seen are needing a way to handle:

- manageiq-ui-service skin - I think I can make this "pluggable" via a rake task in the manageiq-ui-service repo itself which can "find" the directories in the plugins and then symlink the expected values.  This rake task would have to be called during RPM build
- static assets in /public - if a plugin brings static assets in the public directory, this works fine in development mode, but in production mode the assets are behind apache and served from /var/www/miq/vmdb/public directly.  I think this can be handled by an extra step during rpm build to copy into the vmdb/public directory if a plugin brings static assets

simaishi:
Make sure VERSION file update is handled as well.

Fryguy:
I think this one is done for all intents.  Closing now and we can reopen with other issues if we stumble on them.

PR #29, opened 2020-05-01
Use the config gem to simplify options and overrides

Fryguy:
@simaishi Please review.

This PR also brings the ~/OPTIONS directory which can be brought in as a docker volume if needed to override things.  If an options.yml is found there it will override values from config/options.yml

I've also defaulted to ~/OPTIONS and ~/BUILD for simplicity, but allowed those to be overridden with ENV vars even if unlikely.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'awesome_spawn'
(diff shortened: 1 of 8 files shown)


Fryguy:
Part of #2

PR #31, opened 2020-05-01
Productization fix to make build/install work with customized name

simaishi:
Changes needed to make RPM build/install work with customized names.

For now, a few variables in .spec.in and config/options.yml need go be modified. There will be some cleanup coming to make it easier to handle those.


Fryguy:
Part of #2

PR #40, opened 2020-05-05
RPM "Requires" clean up

simaishi:
Moving "Requires" and file contents, based on what need to go to appliance/pods.


simaishi:
@carbonin Can you please review https://github.com/ManageIQ/manageiq-rpm_build/blob/d41ba8a3591be33e745bbe6080a38ea07905bb12/rpm_spec/subpackages/manageiq-system and identify what's NOT needed in pods?

Since pods build currently clones the entire manageiq-appliance repo, I think there is no harm having them in pods for now, but I'd like to move files not needed in pods to manageiq-appliance subpackage.

PR #47, opened 2020-05-15
Always build the SRPM

bdunne:
It can be used to build the binaries locally or through COPR

### DIFF
--- a/lib/manageiq/rpm_build/build_copr.rb
+++ b/lib/manageiq/rpm_build/build_copr.rb
@@ -22,11 +22,14 @@ module ManageIQ
         Dir.chdir(RPM_SPEC_DIR) do
           generate_spec_from_template
 
+          rpm_dir = BUILD_DIR.join("rpms")
+          srpm_dir = rpm_dir.join("src")
+          shell_cmd("rpmbuild -bs --define '_sourcedir #{RPM_SPEC_DIR}' --define '_srcrpmdir #{srpm_dir}' #{rpm_spec}")
+          srpm = srpm_dir.glob("#{OPTIONS.product_name}-*.src.rpm")
           if File.exist?(File.expand_path("~/.config/copr"))
-            shell_cmd("rpmbuild -bs --define '_sourcedir #{RPM_SPEC_DIR}' --define '_srcrpmdir #{RPM_SPEC_DIR}' #{rpm_spec}")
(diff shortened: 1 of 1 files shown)


bdunne:
@miq-bot request_review @simaishi @Fryguy

lib/manageiq/rpm_build/build_copr.rb
```diff
@@ -22,11 +22,14 @@ def generate_rpm
         Dir.chdir(RPM_SPEC_DIR) do
           generate_spec_from_template
 
+          rpm_dir = BUILD_DIR.join("rpms")
+          srpm_dir = rpm_dir.join("src")
```
simaishi, line 26: Would it be a problem to set srpm directory to be outside of rpm dir?  This way, you don't get srpm by default (which doubles the size) and you can separately mount the volume only when you want the srpms.

bdunne: Oh, I was planning to keep the SRPM so that we can rebuild on multiple architectures in the future.  Also for upload to the RPM repository.

simaishi: Right, I understand for our nightly/release build. I was thinking for other 'development' use cases where people most likely wants just RPMs.

lib/manageiq/rpm_build/build_copr.rb
```diff
@@ -22,11 +22,14 @@ def generate_rpm
...
+          shell_cmd("rpmbuild -bs --define '_sourcedir #{RPM_SPEC_DIR}' --define '_srcrpmdir #{srpm_dir}' #{rpm_spec}")
```
chessbyte, line 27: should the `shell_cmd` be replaced with equivalent from `AwesomeSpawn` to prevent injections?

bdunne: We could do that in a separate PR since it is used in many places throughout the code base.

lib/manageiq/rpm_build/build_copr.rb
simaishi: I was thinking about this... since gemset tar file includes ruby gems with compiled binaries, we can't really build RPMs for different architectures from the same source right now...

bdunne: hmm... that's a problem for other arches.  It probably doesn't make sense to publish these "srpms" since they can't be used to build anything else.

PR #48, opened 2020-05-15
Upload built RPMs if credentials are provided

bdunne:
@Fryguy @simaishi Please review

### DIFF
--- a/config/options.yml
+++ b/config/options.yml
@@ -22,3 +22,9 @@ rpm:
   product_url:     https://github.com/ManageIQ/manageiq
 bundler_version:   2.1.4
 npm_registry:
+rpm_repository:
+  s3_api:
+    access_key:
+    secret_key:
+    bucket: rpm-manageiq-org
+    endpoint: nyc3.digitaloceanspaces.com
--- a/lib/manageiq/rpm_build/build_uploader.rb
+++ b/lib/manageiq/rpm_build/build_uploader.rb
@@ -12,7 +12,7 @@ module ManageIQ
 
         require 'digest'
         puts "Uploading files..."
-        BUILD_DIR.glob(File.join('**', '*.rpm')).each do |file|
+        BUILD_DIR.glob(File.join('**', '*.rpm')).sort.each do |file|
           product_directory = release ? OPTIONS[:product_name] : "#{OPTIONS[:product_name]}-nightly"
           destination_name = File.join("builds", product_directory, File.basename(file))
           md5sum = Digest::MD5.file(file).hexdigest
--- a/lib/manageiq/rpm_build/build_uploader.rb
+++ b/lib/manageiq/rpm_build/build_uploader.rb
@@ -1,0 +1,59 @@
+module ManageIQ
+  module RPMBuild
+    class BuildUploader
+      attr_reader :release
+
+      def initialize(options = {})
+        @release = options[:release]
+      end
+
+      def upload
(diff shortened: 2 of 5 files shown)


config/options.yml
```diff
@@ -22,3 +22,9 @@ rpm:
   product_url:     https://github.com/ManageIQ/manageiq
 bundler_version:   2.1.4
 npm_registry:
+rpm_repository:
```
Fryguy, line 25: Should this be under the rpm section?  I ask because this option kind of conflicts with the rpm.repo_name parameter which is for copr.

lib/manageiq/rpm_build/build_uploader.rb
```diff
@@ -0,0 +1,59 @@
...
+      def initialize(options = {})
+        @release = options[:release]
+      end
+
+      def upload
+        return unless OPTIONS.rpm_repository.s3_api.access_key && OPTIONS.rpm_repository.s3_api.secret_key
+
+        require 'digest'
+        puts "Uploading files..."
+        BUILD_DIR.glob(File.join('**', '*.rpm')).each do |file|
```
Fryguy, line 15: I think there be a .sort here after the glob, so there is consistent ordering to the output

lib/manageiq/rpm_build/build_uploader.rb
```diff
@@ -0,0 +1,59 @@
...
+            puts "  Skipping existing file: #{destination_name}"
+          else
+            puts "  Uploading: #{destination_name}"
+            File.open(file, 'rb') do |content|
+              client.put_object(
+                :bucket => OPTIONS.rpm_repository.s3_api.bucket,
+                :key    => destination_name,
+                :body   => content,
+                :acl    => 'public-read'
+              )
```
Fryguy, line 30: Alterantively,

```ruby
client.put_object(
  :bucket => OPTIONS.rpm_repository.s3_api.bucket,
  :key    => destination_name,
  :body   => File.read(file, :mode => "rb"),
  :acl    => 'public-read'
)
```

config/options.yml
bdunne: Does it conflict?

FWIW, I still don't think this belongs here, but I wanted to get something going so that we can get Jansa built.  I think it's strange to put the repo in the rpm (the whole in the part)

lib/manageiq/rpm_build/build_uploader.rb
bdunne: Won't `File.read` read the whole file into memory?  Some of these RPMs can be quite large.  This was based on the documentation for [streaming a file from disk](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#put_object-instance_method)

config/options.yml
Fryguy: I see what you mean, but I don't see it as whole in the part.  To me building RPMs includes optionally deploying them somewhere.  i.e. this repo has never been about just the building part (though the repo name suggests otherwise 😄 )

lib/manageiq/rpm_build/build_uploader.rb
Fryguy: Ah I didn't realize it would stream via the IO.  If so, then yeah, this way is better.

config/options.yml
Fryguy: > Does it conflict?

I guess it doesn't conflict, but it's redundant?

PR #49, opened 2020-05-15
Update rpm repo

bdunne:
DEPENDS ON: #48
- [x] Move common S3 logic to a shared module
- [x] Add code to update the RPM repo
- [ ] Update usage in README
- [x] Testing

### DIFF
--- a/config/options.yml
+++ b/config/options.yml
@@ -28,3 +28,52 @@ rpm_repository:
     secret_key:
     bucket: rpm-manageiq-org
     endpoint: nyc3.digitaloceanspaces.com
+  :arches:
+    - "src"
+    - "noarch"
+    - "x86_64"
+  :content:
+    :10-jansa:
+      :targets:
+        - el8
+      :rpms:
+        :manageiq:
+          - 10
+        :manageiq-release:
+          - 10.0
+        :qpid-proton:
+          - 0.30.0
+        :repmgr10:
+          - 4.0.6
+        :smem:
+          - 1.5
+        :wmi:
+          - 1.3.14
+    :10-jansa-nightly:
+      :targets:
+        - el8
+      :rpms:
+        :manageiq-nightly:
+          - 10
+    :11-kasparov:
+      :targets:
+        - el8
+      :rpms:
+        :manageiq:
+          - 11
+        :manageiq-release:
+          - 11.0
+        :qpid-proton:
+          - 0.30.0
+        :repmgr10:
+          - 4.0.6
+        :smem:
+          - 1.5
+        :wmi:
+          - 1.3.14
+    :11-kasparov-nightly:
+      :targets:
+        - el8
+      :rpms:
+        :manageiq-nightly:
+          - 11
--- a/lib/manageiq/rpm_build/rpm_repo.rb
+++ b/lib/manageiq/rpm_build/rpm_repo.rb
@@ -1,0 +1,73 @@
+module ManageIQ
+  module RPMBuild
+    class RpmRepo
+      include Helper
+      include S3Common
+
+      def update
+        require 'tempfile'
+        Dir.mktmpdir do |tmpdir|
+          work_dir = Pathname.new(tmpdir)
+
+          # Build directory structure /release/{master,jansa}/el8/{noarch,src,x86_64}
+          directories = OPTIONS.rpm_repository.content.flat_map do |branch, values|
+            values[:targets].product(OPTIONS.rpm_repository.arches).map do |target, arch|
+              work_dir.join("release", branch.to_s, target, arch)
+            end
+          end
+          require 'fileutils'
+          directories.each { |dir| FileUtils.mkdir_p(dir) }
+
+          # Fetch the configured RPMs
+          puts "Downloading required RPMs..."
+          OPTIONS.rpm_repository.content.each do |branch, values|
+            values[:rpms]&.each do |rpm, versions|
+              client.list_objects(:bucket => OPTIONS.rpm_repository.s3_api.bucket, :prefix => File.join("builds", rpm.to_s)).contents.each do |object|
+                file = object.key
+                name = File.basename(file)
+                *, target, arch, _rpm = name.split(".")
+                next unless values[:targets].include?(target)
+                next unless versions.any? { |v| name.include?(v.to_s) }
+                cached_rpm = ROOT_DIR.join("rpm_cache", name)
+                unless cached_rpm.file?
+                  puts "Fetching uncached RPM: #{name}"
+                  client.get_object(:bucket => OPTIONS.rpm_repository.s3_api.bucket, :key => file, :response_target => cached_rpm)
+                end
+                destination = work_dir.join("release", branch.to_s, target, arch, name)
+                FileUtils.cp(cached_rpm, destination)
+              end
+            end
+          end
+
+          # puts "Signing release RPMs..."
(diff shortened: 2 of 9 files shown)


lib/manageiq/rpm_build/s3_common.rb
```diff
@@ -0,0 +1,46 @@
...
+              :key    => destination,
+              :body   => content,
+              :acl    => 'public-read'
+            )
+          end
+        end
+      end
+    end
+  end
+end
```
Fryguy, line 46: Missing trailing newline

config/options.yml
```diff
@@ -28,3 +28,40 @@ rpm_repository:
     secret_key:
     bucket: rpm-manageiq-org
     endpoint: nyc3.digitaloceanspaces.com
+  :arches:
+    - "src"
+    - "noarch"
+    - "x86_64"
+  :content:
```
Fryguy, line 35: Is this replaceable with config overrides?

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -0,0 +1,77 @@
...
+              OPTIONS.rpm_repository.arches.each do |arch|
+                directories << work_dir.join("release", branch.to_s, target, arch)
+              end
+            end
+          end
+          directories.each { |dir| FileUtils.mkdir_p(dir) }
+
+          # Fetch the configured RPMs
+          puts "Downloading required RPMs..."
+          require 'rest-client'
```
Fryguy, line 25: I think we are trying to avoid rest-client since it is not being kept up to date.  I am ok for this PR, but we may want to move to just use Net:HTTP afterwards.

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -0,0 +1,77 @@
...
+            values[:rpms]&.each do |rpm, versions|
+              remote_url = File.join("https://#{OPTIONS.rpm_repository.s3_api.bucket}.#{OPTIONS.rpm_repository.s3_api.endpoint}", "?prefix=builds", rpm.to_s)
+              files      = RestClient.get(remote_url).body.scan(/<Key>([^<>]*)<\/Key>/).flatten
+              files.each do |file|
+                name = File.basename(file)
+                *, target, arch, _rpm = name.split(".")
+                next unless values[:targets].include?(target)
+                next unless versions.any? { |v| name.include?(v.to_s) }
+                cached_rpm = ROOT_DIR.join("rpm_cache", name)
+                unless File.file?(cached_rpm)
```
Fryguy, line 36: Minor, but I think you can do `cached_rpm.file?`

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -0,0 +1,77 @@
...
+            end
+          end
+          directories.each { |dir| FileUtils.mkdir_p(dir) }
+
+          # Fetch the configured RPMs
+          puts "Downloading required RPMs..."
+          require 'rest-client'
+          OPTIONS.rpm_repository.content.each do |branch, values|
+            values[:rpms]&.each do |rpm, versions|
+              remote_url = File.join("https://#{OPTIONS.rpm_repository.s3_api.bucket}.#{OPTIONS.rpm_repository.s3_api.endpoint}", "?prefix=builds", rpm.to_s)
```
Fryguy, line 28: Feels like this should be a URI.build instead?

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -0,0 +1,77 @@
...
+              files      = RestClient.get(remote_url).body.scan(/<Key>([^<>]*)<\/Key>/).flatten
```
Fryguy, line 29: As much as we want to avoid XML, I think it's safer to use Nokogiri, and then use some XPath expressions to find what we need

Fryguy: Also, is this XML structure an S3 thing or is this our DO version?  If S3, I'd expect some method in aws-sdk-s3 to handle it.

Fryguy:
@bdunne Personally, I find the YAML config confusing, but I can't think of anything better for now.  Perhaps we can revisit after this first iteration.  Otherwise, this is looking pretty good!

chessbyte:
@bdunne the file `config/options.yml` looks pretty complex, especially with all the version information in there that will likely need to be updated often.  Additionally, I am afraid it may quickly become more complex than https://github.com/ManageIQ/manageiq-release/blob/master/config/labels.yml

Can you lay out the high-level intent around this file?  I want to understand pros/cons of your approach vs other possibilities that we can brainstorm on.

config/options.yml
bdunne: It can be overwritten by other people using it.  It would have to be in sync anywhere that we use it, otherwise it will cause problems.

bdunne:
> @bdunne the file `config/options.yml` looks pretty complex, especially with all the version information in there that will likely need to be updated often. Additionally, I am afraid it may quickly become more complex than https://github.com/ManageIQ/manageiq-release/blob/master/config/labels.yml
> 
> Can you lay out the high-level intent around this file? I want to understand pros/cons of your approach vs other possibilities that we can brainstorm on.

It's intended to be the source of truth for what should live in the RPM repositories.  It only needs to be updated if we want to bring in new packages or new versions of packages.  It will bring in versions that start with the version specified in the config.  With the current config, it will bring in all manageiq builds v10.* to Jansa and all v11.* to Kasparov.  It also includes the `manageiq-release` RPMs for the respective versions and `libssh2`.  I'd also like to bring in the rest of the RPMs from COPR so that we have everything in one place.

There are currently 4 release categories:
- 10-jansa # Release builds only
- 10-jansa-nightly # Nightly builds only
- 11-kasparov # Release builds only
- 11-kasparov-nightly # Nightly builds only

Each of those has 1 build target: # and more in the future
- el8 # CentOS 8

And 3 architectures: # and more in the future
- src # SRPMs
- noarch
- x86_64

config/options.yml
```diff
@@ -22,3 +22,46 @@ rpm:
...
+    :10-jansa:
+      :targets:
+        - el8
+      :rpms:
+        :libssh2:
```
simaishi, line 40: `libssh2` can be dropped as the CentOS repo now contains the package and we don't need to provide our own.

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -0,0 +1,75 @@
...
+          require 'fileutils'
+          directories = []
+          OPTIONS.rpm_repository.content.each do |branch, values|
+            values[:targets].each do |target|
+              OPTIONS.rpm_repository.arches.each do |arch|
+                directories << work_dir.join("release", branch.to_s, target, arch)
+              end
+            end
+          end
+          directories.each { |dir| FileUtils.mkdir_p(dir) }
```
Fryguy, line 21: Super minor, but in a follow up we can clean this up with:

```ruby
all_targets = OPTIONS.rpm_repository.content.values.map { |v| v[:targets] }
directories = all_targets.product(OPTIONS.rpm_repository.arches) do |target, arch|
  work_dir.join("release", branch.to_s, target, arch)
end
```

Fryguy:
This LGTM...I think I'm good to merge as is and we can iterate on things later.

PR #58, opened 2020-06-02
Move ovirt-ansible-roles requires from kickstart to manageiq-gemset

simaishi:
A part of https://github.com/ManageIQ/manageiq-appliance-build/issues/422

Since `python3-ovirt-engine-sdk4` is required by `ovirt-ansible-roles`, I didn't explicitly require that rpm (which was being done in kickstart)

### DIFF
--- a/rpm_spec/subpackages/manageiq-gemset
+++ b/rpm_spec/subpackages/manageiq-gemset
@@ -25,6 +25,9 @@ Requires: cyrus-sasl
 Requires: cyrus-sasl-plain
 Requires: qpid-proton-c
 
+# For RHV
+Requires: ovirt-ansible-roles
+
 # For IMS
 Requires: v2v-conversion-host-ansible
 


rpm_spec/subpackages/manageiq-gemset
```diff
@@ -25,6 +25,9 @@ Requires: cyrus-sasl
 Requires: cyrus-sasl-plain
 Requires: qpid-proton-c
 
+# For RHV
+Requires: ovirt-ansible-roles
+
```
bdunne, line 30: Do we also need `python3-ovirt-engine-sdk4`?

simaishi: ovirt-ansible-roles requires python3-ovirt-engine-sdk4, so no need to explicitly add. Sorry, I should have mentioned in the PR description.

Issue #83, opened 2020-08-21
Support zLinux(s390x) platform


dpkulkarni:
Hello 

We are trying to port `ManageIQ` on zLinux(s390x) architecture. 

I have started to look into the repos to build images to support s390x platform. There are some rpms/packages not available for s390x. 

**Motivation**:
Red Hat Enterprise Linux 8 is already supporting s390x (IBM Z ). We want to port ManageIQ on zLinux architecture

**Issue**: 
Errors during downloading metadata for repository 'manageiq-11-kasparov':

**Logs:**
```
# docker build -t new_rpm:latest -f Dockerfile .
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
STEP 1: FROM registry.access.redhat.com/ubi8/ubi:8.2
STEP 2: ARG ARCH=s390x
--> Using cache 6b3f69e3e53c526132177da92c581b8740b219b4216b53f491a93bf247f6c751
STEP 3: ENV TERM=xterm     APPLIANCE=true     RAILS_USE_MEMORY_STORE=true
--> Using cache 7dbb5f4bf716965b11f2e65e0b7eab275f9ac0be317d58085a03493d73c29949
STEP 4: RUN dnf -y update
--> Using cache c76d579a23c608abff4c5598f005926a1a164a140202d715c056dd8fae35bfb6
STEP 5: RUN curl -L https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo > /etc/yum.repos.d/ansible-runner.repo
--> Using cache 3e18158aac568cba6fe3173533b17b40ed2bb92d143d1b2e378498a5a2f910f7
STEP 6: RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
--> Using cache 21afc3244f3a1b7d40dbafad0f526d5133cc8906bd3e5ec3aab10755ec369059
[...]

PR #84, opened 2020-08-27
Add s390x repo

simaishi:
Wonder if we should add a support for specifying architectures per release?

Related issue: https://github.com/ManageIQ/manageiq-rpm_build/issues/83


simaishi:
qpid-proton is uploaded

PR #92, opened 2020-09-23
Gem cleanup enhancement and refactoring

simaishi:
Also moved contents of scripts/gem_cleanup.sh to generate_gemset.rb. The sh file was originally moved from manageiq-appliance-build repo, but there is no reason this needs to be a standalone shell script as this runs just once during rpm build.

A part of #1


simaishi:
I wanted to remove all unneeded 'ext' directory as well, but I can't figure out a way to programmatically determine which is needed and not.  I've seen 3 different types of gems:

- gem that puts .so to extensions dir and use that (most of the compiled gems)
- gem that doesn't put .so to extensions dir and use what's in ext dir (digest-crc)
- gem that pus .so to extensions and ext dir and use what's in ext dir (sassc)

I can handle the first 2 cases but not sure how to figure out the third case.

Fryguy:
Merging this for now as it looks good...we can deal with the .so files separately (may need to hardcode)

PR #101, opened 2020-10-09
Add the kafka jaas config to kafka.service

jvlcek:
The appliance_console will be used to configure kafka. One of the things the appliance_console will do it to create the file /opt/kafka/config/kafka_server_jaas.conf. 

The /opt/kafka/config/kafka_server_jaas.conf file is introduced to the kafka service with this change.

If the /opt/kafka/config/kafka_server_jaas.conf file does not exist this change will cause the kafka server to fail to start.  This file is required and the appliance_console will need to be used to create it before the service can be started.

### DIFF
--- a/packages/kafka/kafka.service
+++ b/packages/kafka/kafka.service
@@ -3,6 +3,7 @@ Description=kafka service
 After=zookeeper.service
 
 [Service]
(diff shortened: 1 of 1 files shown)


jvlcek:
@miq-bot assign @bdunne 

@bdunne Please review.

bdunne:
If we provide an empty config file, will the server start?

PR #102, opened 2020-10-09
Build RPM for ansible virtualenv

simaishi:
For https://github.com/ManageIQ/manageiq-appliance-build/issues/423

RPM name `manageiq-ansible-venv` and version `1.0.0` need to be confirmed. Since this can be updated independent of manageiq release, I didn't make the version `11.0.0` to make it match.

requirements.txt was created based on  what's currently in manageiq-appliance-build repo. We might want to update, but in a follow up PR.

This basically undo https://github.com/ManageIQ/manageiq-appliance-build/pull/440 and all packages needed are contained in the new RPM.

Also, 
- What need to be productized?  Thinking product name (ManageIQ) for rpm info and rpm prefix is enough. Anything else?
- non-x86_64 arch support. I only tested in x86_64 and some change will be needed for s390x (not sure about ppc64le).

This PR adds files needed to build RPM, but doesn't yet add to product (appliance/container).


packages/manageiq-ansible-venv/README.md
```diff
@@ -0,0 +1,21 @@
+# ManageIQ Ansible Virtualenv RPM Build
+
+Builds RPM for ManageIQ Ansible Virutlalenv in a container image
```
Fryguy, line 3: Minor typo `Virutlalenv` -> `Virtulalenv`

Fryguy (approved):

packages/manageiq-ansible-venv/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ARG ARCH=x86_64
+
+ENV NAME=manageiq-ansible-venv \
+    VERSION=1.0.0 \
+    VENV_ROOT=/var/lib/manageiq \
+    RPM_SPEC=manageiq-ansible-venv.spec
+
+RUN dnf -y --disableplugin=subscription-manager install \
+      http://mirror.centos.org/centos/8.2.2004/BaseOS/${ARCH}/os/Packages/centos-repos-8.2-2.2004.0.1.el8.${ARCH}.rpm \
+      http://mirror.centos.org/centos/8.2.2004/BaseOS/${ARCH}/os/Packages/centos-gpg-keys-8.2-2.2004.0.1.el8.noarch.rpm \
```
Fryguy, line 12: Can we identify which packages need these yum repos, so that if we remove them in the future we know to also remove the repo?

Fryguy:
I agree on the need to productize, but I'd be ok with merging ahead of that and doing it in a follow up PR.

packages/manageiq-ansible-venv/Dockerfile
simaishi: Updated

packages/manageiq-ansible-venv/README.md
simaishi: Oops... fixed.

simaishi:
@seth-priya Can you please try this out on ppc64le?

seth-priya:
@simaishi will do, and get back

seth-priya:
@simaishi the build.sh when run inside the manageiq/ansible_venv_rpm_build container build from
manageiq-rpm_build/packages/manageiq-ansible-venv

using docker build --build-arg ARCH=ppc64le -t manageiq/ansible_venv_rpm_build .

and run using 
docker run -ti manageiq/ansible_venv_rpm_build bash
./build.sh

fails with essentially

    Package 'libffi', required by 'virtual:world', not found
    c/_cffi_backend.c:15:10: fatal error: ffi.h: No such file or directory
     #include <ffi.h>
              ^~~~~~~
    compilation terminated.

on ppc64le, if the steps are correct, let me know if I need to open an issue?

simaishi:
Ah, I think you need to add extra packages for ppc64le. Currently, we have the following code for python modules in manageiq-appliance-build:

```
if [ `uname -m` = "ppc64le" ]; then
  dnf install -y libffi-devel libxslt-devel openssl-devel make;
fi
```
 Looks like you need `libffi-devel` and `openssl-devel` is already installed. We've updated python module versions, so I'm not sure if you still need `libxslt-devel` or not. Would you mind trying out?

simaishi:
@dpkulkarni Can you please try this out on s390x and make necessary changes like ppc64le (#109)?  Thank you.

PR #113, opened 2020-10-27
Update master for Lasker release

simaishi:


config/options.yml
```diff
@@ -37,13 +37,13 @@ rpm_repository:
     - "s390x"
     - "x86_64"
   :content:
-    :11-kasparov:
+    :12-lasker:
```
Fryguy, line 40: Should this replace it or just add a new section?

simaishi: Replace, as we want each branch to track rpms needed for that particular branch/architecture and its versions.

bdunne (approved):

PR #116, opened 2020-11-05
Allow bundler to do the install with the correct build options

Fryguy:
A recent issue was found where these install commands would install the
latest version, and then bundler would resolve to a different version.
Instead of pre-installing, we can let bundler know what options to use
and let it install whatever version is feels is correct.

@simaishi Please review.  I have not tested an actual build with this yet.  Additionally, I'm not sure if this should be --local --global or something else.

### DIFF
--- a/lib/manageiq/rpm_build/generate_gemset.rb
+++ b/lib/manageiq/rpm_build/generate_gemset.rb
@@ -65,8 +65,8 @@ module ManageIQ
           shell_cmd("gem install mime-types -v 2.6.1")
 
           if RUBY_PLATFORM.match?(/powerpc64le|s390x/)
-            shell_cmd("gem install sassc  -- --disable-march-tune-native")
-            shell_cmd("gem install unf_ext -- --with-cxxflags='-fsigned-char'")
+            shell_cmd("bundle config set build.sassc --disable-march-tune-native")
+            shell_cmd("bundle config set build.unf_ext --with-cxxflags='-fsigned-char'")
           end
 
           shell_cmd("bundle config set --local with qpid_proton systemd")


Fryguy:
The specific error that prompted this can be seen here - https://github.com/ManageIQ/manageiq-providers-azure_stack/pull/47#issuecomment-722681796

Basically, before this PR, this line would install unf_ext 0.0.7.7 (the latest) with specific options, but then bundler would resolve to 0.0.7.2 and install it _again_ without the specific options resulting in an error.  This PR defers to bundler to both the resolution _and_ the installation with the options, that way we don't have to keep this line in sync with a specific version number.

(On that note I'm not sure why we install a specific version of mime_types a few lines above instead of just locking down to that version in the Gemfile)

cc @seth-priya (for ppc64le) and @dpkulkarni (for s390x) since both code paths are affected.

simaishi (approved):
I'm not able to run power/s390x buid to confirm this works, but the bundle config setting looks good, so merging.

simaishi:
> (On that note I'm not sure why we install a specific version of mime_types a few lines above instead of just locking down to that version in the Gemfile)

In Gemfile, we use stubbed version of mime-types (https://github.com/ManageIQ/manageiq/pull/14525), but we needed to install the 'standard' version for automate (https://github.com/ManageIQ/manageiq-appliance-build/pull/254).

But, we probably don't need to lock down to that particular version as the rest-client requires mime-types >= 1.16, < 4.0

simaishi:
Kasparov backport details:

```
$ git log -1
commit 14e3af654133e46dd11e59d2bf205cc6bb143c72
Author: Satoe Imaishi <simaishi@redhat.com>
Date:   Mon Nov 9 13:04:23 2020 -0500

    Merge pull request #116 from Fryguy/allow_bundler_to_install

    Allow bundler to do the install with the correct build options

    (cherry picked from commit 0f8f85208c656989608a8493d5aac0439ce12242)
```

seth-priya:
@Fryguy @simaishi thanks for handling this -  confirmed that it build goes well on ppc64le

PR #115, opened 2020-11-04
Ensure updates to server and zookeeper properties are in config-sample

jvlcek:
This change ensures the server.properties and zookeeper.properties files stored in the config-sample directory will include modifications made to these files in this spec.

PR #119, opened 2020-11-12
Bump kafka version after spec changes in #115

bdunne:


bdunne:
cc @jvlcek I already built the RPM with these changes and pushed it to the builds directory.  So, once the nightly build runs the new version should get pulled into the repos.

simaishi (approved):

packages/kafka/kafka.spec
```diff
@@ -107,5 +107,8 @@ rm -rf %{buildroot}
 
 
 %changelog
+* Wed Nov 11 2020 "Brandon Dunne" <bdunne@redhat.com>
```
simaishi, line 110: Can you add \<version\>-\<release\> `2.3.1-2` at the end of this line?

bdunne: :+1: Done

PR #123, opened 2020-11-30
Update the yum repos for s390x

dpkulkarni:
Without the dnf update command for s390x the build fails with the below error:
```
Unable to resolve argument virt:rhel
Error: Problems in request:
missing groups or modules: virt:rhel
The command '/bin/sh -c if [ ${ARCH} != "s390x" ] ; then dnf -y install http://mirror.centos.org/centos/8.2.2004/BaseOS/${ARCH}/os/Packages/centos-repos-8.2-2.2004.0.1.el8.${ARCH}.rpm                                                     http://mirror.centos.org/centos/8.2.2004/BaseOS/${ARCH}/os/Packages/centos-gpg-keys-8.2-2.2004.0.1.el8.noarch.rpm; fi &&     dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm                    https://rpm.manageiq.org/release/12-lasker/el8/noarch/manageiq-release-12.0-1.el8.noarch.rpm &&     dnf -y module enable ruby:2.6 &&     dnf -y module enable nodejs:12 &&     dnf -y module disable virt:rhel &&     dnf -y group install "development tools" &&     dnf config-manager --setopt=epel.exclude=*qpid-proton* --setopt=tsflags=nodocs --save &&     dnf -y install       ansible       cmake       copr-cli       createrepo       glibc-langpack-en       libcurl-devel       libpq-devel       libssh2-devel       libxml2-devel       libxslt-devel       nodejs       openssl-devel       platform-python-devel       postgresql-server       postgresql-server-devel       qpid-proton-c-devel       ruby-devel       rubygem-bundler       sqlite-devel       wget &&     dnf clean all' returned a non-zero code: 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ENV TERM=xterm \
     APPLIANCE=true \
     RAILS_USE_MEMORY_STORE=true
 
+RUN if [ ${ARCH} = "s390x" ] ; then dnf update -y ; fi
 RUN curl -L https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo > /etc/yum.repos.d/ansible-runner.repo
 
 RUN dnf -y --disableplugin=subscription-manager install http://mirror.centos.org/centos/8.2.2004/BaseOS/${ARCH}/os/Packages/centos-repos-8.2-2.2004.0.1.el8.${ARCH}.rpm \


dpkulkarni:
[rpm_build.pdf](https://github.com/ManageIQ/manageiq-rpm_build/files/5627056/rpm_build.pdf)
@simaishi Please find the build log

simaishi:
@dpkulkarni Thank you for the log. Can you also run `docker inspect registry.access.redhat.com/ubi8/ubi:8.2` and send the output?

Your log shows it's using RHEL8 yum repos as well as UBI8 repos:
```
Updating Subscription Management repositories.
Unable to read consumer identity
Subscription Manager is operating in container mode.
Red Hat Enterprise Linux 8 for IBM z Systems - AppStream (RPMs) 5.3 MB/s | 19 MB 00:03
Red Hat Enterprise Linux 8 for IBM z Systems - BaseOS (RPMs)7.9 MB/s | 19 MB 00:02
Red Hat Universal Base Image 8 (RPMs) - BaseOS 591 kB/s | 557 kB 00:00
Red Hat Universal Base Image 8 (RPMs) - AppStream 4.3 MB/s | 4.5 MB 00:01
Red Hat Universal Base Image 8 (RPMs) - CodeReady Builder 23 kB/s | 11 kB 00:00
```

When using UBI container images, you should not be seeing RHEL 8 repos. On my x86_64 build, this is what I get:

```
Updating Subscription Management repositories.
Unable to read consumer identity
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Red Hat Universal Base Image 8 (RPMs) - BaseOS  959 kB/s | 771 kB     00:00
Red Hat Universal Base Image 8 (RPMs) - AppStre 6.8 MB/s | 4.8 MB     00:00
Red Hat Universal Base Image 8 (RPMs) - CodeRea  18 kB/s |  13 kB     00:00
```

So it looks like your base image is registered. Running `docker inspect` on that base image will tell exactly which image is being used here as a starting point.

Since RHEL8 repos are in place, `dnf update` is taking rpms from the RHEL8 repos, not from the UBI8 repos. UBI8 repos contain only subset of packages available in RHEL8 repos, so this test needs to be re-run using only UBI8 repos.

Lastly, it would be great if you can post the log in non-PDF format. That wraps the text and it's very hard to read... Thanks!

simaishi:
@dpkulkarni I looked at why we added `dnf -y module disable virt:rhel` initially, and that was because you ran into issue on s390x (https://github.com/ManageIQ/manageiq-rpm_build/issues/83#issuecomment-702811563).

Looking at the log now, I realized that build is also using RHEL8 repos. My guess is that if you only had UBI8 repos at that time, you would have not seen that issue (but you might have seen other issue....)

So... my suggestion is 1) first make sure RHEL8 repos are NOT used, and 2) remove `dnf -y module disable virt:rhel` from Dockerfile and run the build. You might see some issues, but that's the real issue we need to fix.

dpkulkarni:
@simaishi Without adding RHEL8 repos and removing `virt:rhel`,  "development tools' is not available with UBI8.2
```
Module or Group 'development tools' is not available.
Error: Nothing to do.
```
Please find the log.

[rpm_build_ubi_repo_only copy.docx](https://github.com/ManageIQ/manageiq-rpm_build/files/5637126/rpm_build_ubi_repo_only.copy.docx)

simaishi:
Then, that's probably coming from CentOS, which isn't available for s390x. We have to figure out exactly which packages from "Development Tools" is needed and list individual packages rather than trying to group install in that case.

I can play with that on x86_64 and come up with the list. If you've always used RHEL8 repos in your testing, it's possible some other packages could be missing without RHEL8 repo. So you might want to try running the build (without "development tools") and see if anything else fails in the meantime.

dpkulkarni:
@simaishi I just removed the "development tools" and ran the build and got the errors with other packages as well.
The build is not yet successful. 

attaching.
[More_packages_remove.docx](https://github.com/ManageIQ/manageiq-rpm_build/files/5637563/More_packages_remove.docx)

simaishi:
@dpkulkarni not having postgresql is a problem... where do you usuall get postgresql from for s390x?

dpkulkarni:
postgresql image for s390x is available on Dockerhub

simaishi:
@dpkulkarni We're not looking for a postgresql container image, but looking for RPM that can run on UBI8/CentOS8. On x86_64 and ppc64le, that comes from CentOS AppStream. But there is no s390x availab for CentOS.

This isn't the first time we ran to this issue. Since we

dpkulkarni:
@simaishi 

> @dpkulkarni We're not looking for a postgresql container image, but looking for RPM that can run on UBI8/CentOS8. On x86_64 and ppc64le, that comes from CentOS AppStream. But there is no s390x available for CentOS.
> 
> This isn't the first time we ran into this issue. We rely on UBI/CentOS to provide tools we need, to run our app. So I'm not sure what the expectation should be when trying to make this work on an architecture that's not supported by them... cc @Fryguy @chessbyte

postgresql rpm's are available for s390x which we can use

dpkulkarni:
> Then, that's probably coming from CentOS, which isn't available for s390x. We have to figure out exactly which packages from "Development Tools" is needed and list individual packages rather than trying to group install in that case.
> 
> I can play with that on x86_64 and come up with the list. If you've always used RHEL8 repos in your testing, it's possible some other packages could be missing without RHEL8 repo. So you might want to try running the build (without "development tools") and see if anything else fails in the meantime.
@simaishi 
If you got some time to get the list which are required so that I can see if I can get them on s390x or can build and include for s390x and proceed further.. Thanks

simaishi:
Where are they available? Your log from https://github.com/ManageIQ/manageiq-rpm_build/pull/123#issuecomment-738131930 shows:

```
No match for argument: createrepo
No match for argument: postgresql-server
No match for argument: postgresql-server-devel
```

dpkulkarni:
I have not included them yet in the build. 
These are the RPMs
```
https://downloads.redhat.com/pub/redhat/rhel/rhel-8-beta/appstream/s390x/Packages/postgresql-10.5-1.el8+1546+27ad5f8e.s390x.rpm
https://downloads.redhat.com/pub/redhat/rhel/rhel-8-beta/appstream/s390x/Packages/postgresql-server-10.5-1.el8+1546+27ad5f8e.s390x.rpm
```

simaishi:
So the issue is... we can't use RPMs from RHEL8 repos. RHEL and UBI are under different licenses and only what's available under UBI can be freely distributed...

dpkulkarni:
@simaishi 
The components which we are not getting from UBI have to be built from the source code for s390x and we have built the postgresql for s390x using the source
What do you suggest?

simaishi:
@dpkulkarni The changes needed to run build without "development tools": https://github.com/simaishi/manageiq-rpm_build/commit/b79b67a8b0993ebf0986f35277ee6c865007fa72. I haven't fully tested yet, but this should be a good start for you.

We still need to figure out what to do with rpms not available by default as they need to be publically available.

dpkulkarni:
@simaishi 
Tried building the components mentioned in the https://github.com/simaishi/manageiq-rpm_build/commit/b79b67a8b0993ebf0986f35277ee6c865007fa72 but some of them are not available on ubi8.2 but are present in ubi7.
One alternative to go ahead..
For s390x shall we have a separate Dockerfile (Dockerfile.s390x) and have the packages from ubi7 and build the images?

simaishi:
@dpkulkarni I don't recommend using ubi7 just for s390x. We only test/build each branch for one UBI/CentOS version and there is no gurantee things will work as expected on other versions. Also, we might need to make version specific changes in the future and that will make s390x even more apart from the rest of the product and make things harder to maintain.

dpkulkarni:
@simaishi 
Thanks for the inputs .. for now I am closing this PR as no change since we have to come up a way with rpms not available by default as they need to be publicly available.  

Opening an issue for the same to discuss.

PR #126, opened 2020-12-16
Add librdkafka RPM

jvlcek:
This is part of the solution for:
ManageIQ/manageiq-appliance-build#456

When gem `rdkafka` is installed/built SSL support will not be built in unless the librdkafka RPM is installed.

This PR will add the librdkafka RPM which will ensure the rdkafka gem is installed with SSL support.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN if [ ${ARCH} != "s390x" ] ; then dnf -y install http://mirror.centos.org/cen
       glibc-langpack-en \
       libcurl-devel \
       libpq-devel \
+      librdkafka \
       libssh2-devel \
       libxml2-devel \
       libxslt-devel \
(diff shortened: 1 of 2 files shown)


jvlcek:
@miq-bot add_label enhancement
@miq-bot assign @simaishi

jvlcek:
I discussed where best to specify the librdkafka RPM requirement with @simaishi, who suggested to put it where this PR does.

@agrare Please take a look.

agrare (approved):

simaishi:
@jvlcek just realized reading the PR description. If the rpm is needed when installing the gems, you need to add this rpm to Dockerfile. If the rpms is needed at run time, you need to add to the rpm spec file.

simaishi:
-devel package is available in PowerTools repo (not enabled by default), if needed. Since @jvlcek tested without -devel and it works, I guess that's not needed though....

Issue #128, opened 2021-01-05
Flushing cache for yum repo sometimes fails, causing build script to end prematuarlly


simaishi:
```
Cleaning release/11-kasparov-nightly:
Purging the cache for files that were uploaded
/root/.gem/ruby/gems/droplet_kit-3.12.0/lib/droplet_kit/error_handling_resourcable.rb:14:in `block (2 levels) in included': 500: {"id": "Internal Server Error", "message": "Server Error" } (DropletKit::Error)
        from /root/.gem/ruby/gems/resource_kit-0.1.7/lib/resource_kit/action_invoker.rb:19:in `instance_exec'
        from /root/.gem/ruby/gems/resource_kit-0.1.7/lib/resource_kit/action_invoker.rb:19:in `handle_response'
        from /root/.gem/ruby/gems/resource_kit-0.1.7/lib/resource_kit/action_invoker.rb:14:in `call'
        from /root/.gem/ruby/gems/resource_kit-0.1.7/lib/resource_kit/method_factory.rb:16:in `block in method_for_action'
        from /build_scripts/lib/manageiq/rpm_build/rpm_repo.rb:82:in `block in update'
        from /usr/share/ruby/tmpdir.rb:93:in `mktmpdir'
        from /build_scripts/lib/manageiq/rpm_build/rpm_repo.rb:9:in `update'
        from bin/build.rb:52:in `<main>'
Command exited with non-zero status 1
```

simaishi:
cc @bdunne

PR #129, opened 2021-01-05
Retry flushing the CDN cache if there is an error

bdunne:
I can't reproduce this problem locally, so I don't know what causes it.

Fixes: https://github.com/ManageIQ/manageiq-rpm_build/issues/128

### DIFF
--- a/lib/manageiq/rpm_build/rpm_repo.rb
+++ b/lib/manageiq/rpm_build/rpm_repo.rb
@@ -1,3 +1,5 @@
+require 'droplet_kit'
+
 module ManageIQ
   module RPMBuild
     class RpmRepo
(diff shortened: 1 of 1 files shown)


lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -74,15 +76,20 @@ def update
...
+
+      private
+
+      def flush_cdn_cache(uploaded_files)
+        puts "Purging the cache for files that were uploaded"
+        digitalocean_client = DropletKit::Client.new(:access_token => OPTIONS.rpm_repository.digitalocean_access_token)
+        cdn_id = digitalocean_client.cdns.all.detect { |i| i.origin == "#{OPTIONS.rpm_repository.s3_api.bucket}.#{OPTIONS.rpm_repository.s3_api.endpoint}" }.id
+        digitalocean_client.cdns.flush_cache(:id => cdn_id, :files => uploaded_files)
+      rescue DropletKit::Error
+        retry
```
Fryguy, line 91: This will continually retry forever...should we put a cap on it, or only let it try once?

PR #131, opened 2021-01-08
Fix rpm 'release' floating addition

simaishi:
Didn't realize 0.2 + 0.1 isn't 0.3, but it's 0.30000000000000004...


Fryguy:
Floating Point Math is fun!

PR #141, opened 2021-03-25
Change the systemd unit file location to ./systemd/

agrare:
Some filesystems were having issues with a ./build/ directory and a ./BUILD file.  Move systemd unit files to a ./systemd/ directory

https://github.com/ManageIQ/manageiq/pull/21122


bdunne:
@agrare should this be `morphy/yes?` as well?

agrare:
See https://github.com/ManageIQ/manageiq-rpm_build/pull/134#issuecomment-916295187

PR #146, opened 2021-04-09
[WIP] remove watchdog

kbrock:
part of a few PRs:

- https://github.com/ManageIQ/manageiq/pull/21146
- https://github.com/ManageIQ/manageiq-appliance/pull/311

don't run evmserver `update_stop` in rpm update. running `stop` instead.

we are getting out of the business of running a watchdog service

### DIFF
--- a/rpm_spec/subpackages/manageiq-appliance
+++ b/rpm_spec/subpackages/manageiq-appliance
@@ -61,7 +61,7 @@ if posix.access('/bin/evmserver.sh', 'x') then
   if ( pid == -1 ) then
     print ("The fork failed.")
   elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_stop')
+    posix.exec('/bin/evmserver.sh', 'stop')
   else
     posix.wait(pid)
   end
@@ -73,7 +73,7 @@ if posix.access('/bin/evmserver.sh', 'x') then
   if ( pid == -1 ) then
     print ("The fork failed.")
   elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_start')
+    posix.exec('/bin/evmserver.sh', 'start')
   else
     posix.wait(pid)
   end


kbrock:
> @bdunne Can the rpm calls just call systemctl stop evmserverd? I think it's important that the rpm updates work as expected, but I don't think it needs to go through all of that ceremony once the watchdog is gone.

>As an aside, can you give more insight into what the purpose of the watchdog was?
> -- @Fryguy 

backwards compatibility does suggest stopping and starting.
but a typical rpm doesn't do stuff like this - does it?

I expect to manually stop the service, upgrade the rpm, then migrate the database and do some other work before manually starting up the server again. But I guess every admin could expect something different.

Fryguy:
> but a typical rpm doesn't do stuff like this - does it?

I don't know and it's a good question.  If it's not common for an rpm to restart its own services, then I'm fine with not including that.  @bdunne?

agrare:
I'm pretty sure it is common for upgrading an rpm to restart the associated services [citation needed 😆 ]

kbrock:
ok, I left in.

I just remembered why I wanted to remove them.

We had a customer that was not ready to deploy yet but upgraded rpms on 30 of their worker appliances.

This kicked off evmserver on those appliances and started populating server records in the database with invalid `v2_keys`. It was not too hard to blow away the records and get everything back up and running, but it sure caused a few hours of havoc until we realized that a few dozen machines had something running and had that is why invalid rows were in the `miq_servers` table.

rpm_spec/subpackages/manageiq-appliance
```diff
@@ -73,7 +73,7 @@ if posix.access('/bin/evmserver.sh', 'x') then
   if ( pid == -1 ) then
     print ("The fork failed.")
   elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_start')
+    posix.exec('/bin/evmserver.sh', 'start')
```
bdunne, line 76: I think this will cause problems during appliance image build.  In the anaconda environment it will try to start the server, right?

bdunne: The difference is that `update_start` only starts the app if it was stopped with `update_stop`.

kbrock: so should we touch a file if we stop the service, so it will try and start it?

For regular upgrades, I don't want it to start the service either. A few times a customer has said that it started the service when they did not want it to start after upgrading their rpms

kbrock:
WIP: currently there are a few outstanding features that we need to resolve before removing this

the rpm upgrade process properly does not start this process if it was not originally running (do need to validate that this properly works)
if an appliance starts up, when the database is down (possibly also starting up at the same time) then this will wait a little and try again. It will also determine if it is in the middle of an upgrade.

the multi appliance process still had a bit of discussion. Instead of tackling this now, it will be addressed separately (and after) root services are addressed

kbrock:
WIP: see https://github.com/ManageIQ/manageiq-rpm_build/pull/146

PR #152, opened 2021-05-26
add common usage examples

kbrock:
It was not clear before how these commands fit together.

Here is how I understood how our process worked.
Don't think I dumbed it down too much but please let me know

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,6 @@ repos:
 
 The option file is brought into docker by mounting the directory with `-v $(pwd)/OPTIONS:/root/OPTIONS`
 
-Alternatively, a static directory may be used like `/opt/OPTIONS/options.yml`.
-
 ### Using a custom NPM registry
 
 Any of the keys in [options.yml](blob/master/config/options.yml) can be overridden.
@@ -109,47 +107,20 @@ The purpose of this process is to build code from `manageiq-core`, `manageiq-gem
 - `/root/BUILD/rpm_spec/*.tar.gz`
 - `/root/BUILD/rpms/x86_64/manageiq-*.rpm`
 
-Depending upon your use case, there are a few ways to obtain these artifacts:
-
-- Have the container put the artifacts into a mounted volume.
-- Manually copy the artifacts to a mounted volume.
-- Extracting the artifacts from the docker container with a docker command.
-
-### Having the container build files into a mounted volume
-
-Since the files are not uploaded, they need to be retrieved from the repo before the container exits.
-
-For this example, 2 directories exist in the current directory (`pwd`):
-
-- `OPTIONS` a directory contains the `options.yml` file
-- `BUILD` a directory that will contain the `tar.gz` and `rpm` files.
-
-Note: The contents of `BUILD` will be deleted before each run.
+One approach is to build and mount a volume that will receive all artifacts:
 
 ```sh
 docker run --rm -v `pwd`/OPTIONS:/root/OPTIONS -v `pwd`/BUILD:/root/BUILD manageiq/rpm_build:latest build
 ```
 
-### Manually copying files to a mounted volume
-
-The commands to setup a build in the docker container can be found in the [entrypoint](blob/master/container-assets/user-entrypoint.sh#L17) script.
-
-The process has a few steps:
-
-- start up a build environment
-- run the build
-- copy the assets out
-- tell the build environment to shut down
+Another approach is to build and copy out the desired artifacts:
 
 ```sh
-# setup docker and remember the CONTAINER id
 CONTAINER='my-build-container'
-docker run --name ${CONTAINER} -d -it -v $(pwd)/OPTIONS:/root/OPTIONS manageiq/rpm_build:latest
-docker exec -it ${CONTAINER} /build_scripts/container-assets/user-entrypoint.sh build
+docker run --name ${CONTAINER} -v `pwd`/OPTIONS:/root/OPTIONS manageiq/rpm_build:latest build
 docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
-docker attach ${CONTAINER} # type 'exit'
+docker rm ${CONTAINER}
 ```
-
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -1,21 +1,100 @@
 # ManageIQ RPM Build
 
-This repository contains code to build RPMs for ManageIQ appliances and container images:
+## Summary
+
+This repository contains code to build RPMs for ManageIQ appliances and container images.
+
+It performs the following tasks:
 
   * Clone source repos
   * Run bundle install, compiles assets for UI and a few other tasks needed
   * Create tarballs
   * Build RPMs (locally in the container image or in Copr)
 
+## Output
+
+It produces the following rpms:
+
+| rpm                        | contents                                                          |
+|----------------------------|-------------------------------------------------------------------|
+| `manageiq-appliance-tools` | **empty**                                                         |
+| `manageiq-gemset`          | `/opt/manageiq/manageiq-gemset/` gems, providers, ui-classic      |
+| `manageiq-gemset-services` | `manageiq-provider-*{.target,@.service}`                          |
+| `manageiq-core`            | `vmdb/{app,certs,content,lib,product,systemd}`                    |
+| `manageiq-core-services`   | `manageiq-*{.target,@.service}`                                   |
+| `manageiq-system`          | `/etc/{default,chron}, /usr/bin/evm*, system/{evm,miq}*.service`  |
+| `manageiq-appliance`       | `/etc/httpd/conf.d/manageiq-*, /etc/motd.manageiq`                |
+| `manageiq-ui`              | `vmdb/public/{assets,packs,ui}`                                   |
+| `manageiq-pods`            | **empty**                                                         |
+
+* `/usr/lib/systemd/system/` contains systemd service files.
+* `/var/www/miq/vmdb` contains application files.
+* `/opt/manageiq/manageiq-appliance` contains files linked into `/etc/`.
+* `/opt/manageiq/manifest` contains manifests.
+
 ## Steps
 
-1. Build container image
+## Obtain container image for building RPMs
 
-2. Prepare configuration
+You can either download the latest build image or building a new one locally:
 
-   - Create a local directory with an `options.yml` with the configuration changes
-     you want to make, then when running the container image, mount that directory
-     with `-v <dir>:/root/OPTIONS`.
+  - `docker pull manageiq/rpm_build:latest`
+  - `docker pull manageiq/rpm_build:latest-jansa`
+  - `docker build --pull --tag $USER/rpm_build:latest .`
+
+Typically the first example works best but if you are modifying which files end up in the rpm, the third example is the one you want.
+
+## Building a release
+
+The github tag/branch is passed on the command line e.g.: `--git-ref jansa`. A few other options and explanations are availabe in [build.rb](blob/master/bin/build.rb#L8) for the basic installs.
+
+
+```sh
+docker run --rm manageiq/rpm_build:latest build --git-ref lasker --update-rpm-repo
+```
+
+For older builds, the `git-ref` is already stored in `config/options.yml` and not necessary.
+
+```sh
+docker run --rm manageiq/rpm_build:jansa-3 build --update-rpm-repo
+```
+
+## Building a custom release
+
+Sometimes it is necessary to build rpms with code from pull requests. This explains how to do that.
+
+### Defining build parameters
+
+The [options.yml](blob/master/config/options.yml) points to the appropriate repos and tag/branches.
+
+A typical `options.yml` will only override the source and the refs. This example uses branch `feature1` from kbrock's github repo:
+
+```sh
+mkdir OPTIONS
+vi OPTIONS/options.yml
+```
+
+```yaml
+---
+product_name:      manageiq
+repos:
+  ref:             master
+  manageiq:
+    url:           https://github.com/kbrock/manageiq.git
+    ref:           feature1
+  manageiq_appliance:
+    url:           https://github.com/kbrock/manageiq-appliance.git
+    ref:           feature1
+```
+
+The option file is brought into docker by mounting the directory with `-v $(pwd)/OPTIONS:/root/OPTIONS`
+
+Alternatively, a static directory may be used like `/opt/OPTIONS/options.yml`.
+
+### Using a custom NPM registry
+
+Any of the keys in [options.yml](blob/master/config/options.yml) can be overridden.
+
    - If overriding the NPM registry, set the `npm_registry` key in the `options.yml`.
    - If building RPMs in Copr,
      - set the `rpm.repo_name` key in the `options.yml`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,58 +1,195 @@
...
+* `/var/www/miq/vmdb` contains vmdb files.
+* `/opt/manageiq/manageiq-appliance` contains files linked into `/etc/`.
+* `/opt/manageiq/manifest` contains manifests.
+
 ## Steps
 
-1. Build container image
+## Obtain container base image
+
+You can either downloading the latest build image or building a new one locally:
```
Fryguy, line 39:
```suggestion
You can either downloading the latest build image or build a new one locally:
```

README.md
```diff
@@ -1,58 +1,195 @@
...
+
+* `/usr/lib/systemd/system/` contains systemd service files.
+* `/var/www/miq/vmdb` contains vmdb files.
+* `/opt/manageiq/manageiq-appliance` contains files linked into `/etc/`.
+* `/opt/manageiq/manifest` contains manifests.
+
 ## Steps
 
-1. Build container image
+## Obtain container base image
```
Fryguy, line 37: This seems unclear...instead can we use

```suggestion
## Obtain container image for building RPMs
```

README.md
```diff
@@ -1,58 +1,195 @@
...
+
+If you modify files in in this respository like adding a file to the rpm, it is necessary to rebuild this image as shown by the third example.
+Do keep in mind that having a local `Gemfile.lock` before running `docker build` can cause versioning issues with bundler.
+
+## Building a release
+
+The github tag/branch is passed on the command line e.g.: `--git-ref jansa`. A few other options and explanations are availabe in [build.rb](blob/master/bin/build.rb#L8) for the basic installs.
+
+```sh
+docker run -v --rm manageiq/rpm_build:latest-jansa build --git-ref jansa --update-rpm-repo
```
Fryguy, line 53: This has a -v but doesn't pass an actual volume...was it missed?  (Same with the next command)

README.md
```diff
@@ -1,58 +1,195 @@
...
+docker run -v --rm manageiq/rpm_build:jansa build --build-type release --git-ref jansa-1-alpha1 --update-rpm-repo
+```
+
+## Building a custom release
+
+Sometimes it is necessary to build rpms with code from pull requests. This explains how to do that.
+
+### Defining build parameters
+
+The [options.yml](blob/master/config/options.yml) pointing to the approperate repos and tag/branches.
```
Fryguy, line 66:
```suggestion
The [options.yml](blob/master/config/options.yml) points to the appropriate repos and tag/branches.
```

README.md
```diff
@@ -1,58 +1,195 @@
...
+### Having the container build files into a mounted volume
 
-Alternatively, the build can be started directly by passing `build` to the
-`docker run` command:
+Since the files are not uploaded, they need to be retrieved out of the repo before the container exits.
 
+Note: all mounted volumes (i.e.: using the `-v` option) need to be defined in docker's preferences:
+
+```
+Settings > Resources > FILE SHARING
```
Fryguy, line 123: What is this referencing?

README.md
```diff
@@ -1,58 +1,195 @@
...
-     ```
-     bin/build.rb --build-type release --git-ref jansa-1-alpha1 --update-rpm-repo
-     ```
+### Having the container build files into a mounted volume
 
-Alternatively, the build can be started directly by passing `build` to the
-`docker run` command:
+Since the files are not uploaded, they need to be retrieved out of the repo before the container exits.
 
+Note: all mounted volumes (i.e.: using the `-v` option) need to be defined in docker's preferences:
```
Fryguy, line 120: I don't understand what this means - I've never needed the volumes in docker preferences as far as I can remember.

README.md
```diff
@@ -1,58 +1,195 @@
...
+
+For this example, 2 directories exist in the current directory (`pwd`):
+
+- `OPTIONS` a directory contains the `options.yml` file
+- `BUILD` a directory that will contain the `tar.gz` and `rpm` files.
+
+Note: The contents of `BUILD` will be deleted before each run.
+
+```sh
+docker run -v --rm -v `pwd`/OPTIONS:/root/OPTIONS -v `pwd`/BUILD:/root/BUILD manageiq/rpm_build:latest build
```
Fryguy, line 134: There's an extra -v after the run.

README.md
```diff
@@ -1,58 +1,195 @@
...
+### Manually copying files to a mounted volume
+
+Since defining the mount points can be inconvenient, it is possible to just run a build and copy out the files out.
+The trick is to run the container as a daemon and in a way that it does not exit before you have copied out the resources.
+
+#### Running the container in daemon mode
+
+```sh
+# setup docker and remember the CONTAINER id
+CONTAINER=`docker run -d -it -v $(pwd)/OPTIONS:/root/OPTIONS manageiq/rpm_build:latest`
```
Fryguy, line 146: it's generally a lot easier if you're running a daemonized container to give it a name using `--name`.  Then later you just reference the container by name directly, and then no need for env vars or anything.

README.md
```diff
@@ -1,58 +1,195 @@
...
+cp -v /root/BUILD/rpms/x86_64/m* /root/OPTIONS/rpms/
 ```
-docker run -it <miq-rpm_build image> build [--build-type <type>] [--git-ref <ref>] [--update-rpm-repo]
+
+#### Pulling the build artifacts out of docker
+
+An alternative is to exit out of docker and have docker go into the container and copy them to your local file system
+
+```sh
+docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
```
Fryguy, line 182: This is generally the easiest way if you're using a daemonized container...I think we should just remove the "Pushing the build artifacts to a mounted volume" section, personally, to simplify (too many choices => paradox of choice)

Fryguy (approved):
Overall, LGTM...would like @bdunne to also review.

README.md
```diff
@@ -1,58 +1,195 @@
...
+| `manageiq-gemset-services` | `manageiq-provider-*{.target,@.service}`                          |
+| `manageiq-core`            | `vmdb/{app,certs,content,lib,product,systemd}`                    |
+| `manageiq-core-services`   | `manageiq-*{.target,@.service}`                                   |
+| `manageiq-system`          | `/etc/{default,chron}, /usr/bin/evm*, system/{evm,miq}*.service`  |
+| `manageiq-appliance`       | `/etc/httpd/conf.d/manageiq-*, /etc/motd.manageiq`                |
+| `manageiq-ui`              | `vmdb/public/{assets,packs,ui}`                                   |
+| `manageiq-pods`            | **empty**                                                         |
+
+* `/usr/lib/systemd/system/` contains systemd service files.
+* `/var/www/miq/vmdb` contains vmdb files.
```
bdunne, line 31:
```suggestion
* `/var/www/miq/vmdb` contains application files.
```

README.md
```diff
@@ -1,58 +1,195 @@
...
+## Obtain container base image
+
+You can either downloading the latest build image or building a new one locally:
+
+  - `docker pull manageiq/rpm_build:latest`
+  - `docker pull manageiq/rpm_build:latest-jansa`
+  - `docker build --pull --tag $USER/rpm_build:latest .`
+
+If you modify files in in this respository like adding a file to the rpm, it is necessary to rebuild this image as shown by the third example.
+Do keep in mind that having a local `Gemfile.lock` before running `docker build` can cause versioning issues with bundler.
```
bdunne, line 46: Should we dockerignore Gemfile.lock?

README.md
bdunne: Is this a Docker on Mac thing?

README.md
kbrock: if you want to add a mounted volume using `-v`, then you will need to add the volume to your file share and restart docker.

README.md
kbrock: yea mac thing

README.md
kbrock: to be honest, I tend to go in, build it, an while I'm in there just copy to /root/OPTIONS

I'd prefer just a `--copy-rpms $TARGET` option using build and be done with it.

bdunne: > to be honest, I tend to go in, build it, an while I'm in there just copy to /root/OPTIONS
> 
> I'd prefer just a `--copy-rpms $TARGET` option using build and be done with it.

Wouldn't it be easier to use a second volume mount rather than daemonize, exec into it, then copy to a volume mount?

README.md
Fryguy: Maybe?  In production builds we _want_ to bring a Gemfile.lock during rpm build because we want to ensure we aren't arbitrarily updating gems.  We usually have a Gemfile.lock.release file, but I don't know how @simaishi made that be the "real" lock file.

README.md
kbrock: thnx - verified - removed

README.md
kbrock: 1. removing push
2. I find it easier to mount all these directories but I will try again and see how it goes. I've already been convinced of a number of changes that are making this easier

README.md
kbrock: thanks for solving this. With this knowledge, the `-v` option is much better

kbrock:
@bdunne Could you double check my usage for building jansa above (or we could update to k or l - up to you)

How do you build an old release? Do you use `rpm_build:latest` or `latest-$release`, or another? And which `--git-ref` do you use?  Maybe `jansa1`?

```sh
docker run --rm manageiq/rpm_build:latest-jansa build --git-ref jansa1 --update-rpm-repo
```

bdunne:
Yes, for older branches we use `latest-jansa`.  I don't think we need the `--git-ref` for the older branches since it's set [here](https://github.com/ManageIQ/manageiq-rpm_build/blob/lasker/config/options.yml#L4).

README.md
kbrock: thanks for the question. Misunderstandings of previous errors had me going in a tangent.

README.md
Fryguy: Oh I misunderstood this - I was thinking the Gemfile.lock was the _application's_ Gemfile.lock as opposed to the one for manageiq-rpm_build itself.  Yeah, for the latter, we can probably .dockerignore that.

kbrock:
ok, updated the statement around `--git-ref` not being for tagged builds.

I'm good to go with this one if others are as well.

README.md
```diff
@@ -1,58 +1,161 @@
...
      Any RPMs not in the cache will be downloaded to the cache first.
 
-3. Start the container image
+### Artifacts
+
+The purpose of this process is to build code from `manageiq-core`, `manageiq-gemset`, and `manageiq-appliance` into 2 sets of artifacts:
 
-   `docker run -it <image>`
+- `/root/BUILD/rpm_spec/*.tar.gz`
+- `/root/BUILD/rpms/x86_64/manageis-*.rpm`
```
Fryguy, line 110:
```suggestion
- `/root/BUILD/rpms/x86_64/manageiq-*.rpm`
```

README.md
```diff
@@ -1,58 +1,161 @@
...
 
-Alternatively, the build can be started directly by passing `build` to the
-`docker run` command:
+Since the files are not uploaded, they need to be retrieved out of the repo before the container exits.
+
+Note: all mounted volumes (i.e.: using the `-v` option) need to be defined in docker's preferences:
 
 ```
-docker run -it <miq-rpm_build image> build [--build-type <type>] [--git-ref <ref>] [--update-rpm-repo]
+Settings > Resources > FILE SHARING
```
Fryguy, line 125: We probably should mark this as being Docker on Mac specific...this steps doesn't make sense for normal Docker.  I'm ok with a follow-up PR though.

Fryguy:
LGTM - one typo.  I'll let @bdunne merge.

README.md
kbrock: thanks - I'll dump this section. now that you found out that you do not need to put every directory into there, this makes the use case much easier

README.md
```diff
@@ -1,58 +1,155 @@
...
+## Building a release
+
+The github tag/branch is passed on the command line e.g.: `--git-ref jansa`. A few other options and explanations are availabe in [build.rb](blob/master/bin/build.rb#L8) for the basic installs.
+
+
+```sh
+docker run --rm manageiq/rpm_build:latest build --git-ref lasker --update-rpm-repo
+```
+
+For older older builds, the `git-ref` is already stored in `config/options.yml` and not necessary.
```
bdunne, line 56:
```suggestion
For older builds, the `git-ref` is already stored in `config/options.yml` and not necessary.
```

README.md
```diff
@@ -1,58 +1,155 @@
...
+    url:           https://github.com/kbrock/manageiq.git
+    ref:           feature1
+  manageiq_appliance:
+    url:           https://github.com/kbrock/manageiq-appliance.git
+    ref:           feature1
+```
+
+The option file is brought into docker by mounting the directory with `-v $(pwd)/OPTIONS:/root/OPTIONS`
+
+Alternatively, a static directory may used like `/opt/OPTIONS/options.yml`.
```
bdunne, line 92:
```suggestion
Alternatively, a static directory may be used like `/opt/OPTIONS/options.yml`.
```

README.md
```diff
@@ -1,58 +1,155 @@
...
 
-   `docker run -it <image>`
+The purpose of this process is to build code from `manageiq-core`, `manageiq-gemset`, and `manageiq-appliance` into 2 sets of artifacts:
 
-4. Run the script
+- `/root/BUILD/rpm_spec/*.tar.gz`
+- `/root/BUILD/rpms/x86_64/manageis-*.rpm`
 
-   - Run `bin/build.rb`, optionally overriding git ref (branch or tag) and release type
+Depending upon your usecase, there are a few ways to obtain these artifacts:
```
bdunne, line 112:
```suggestion
Depending upon your use case, there are a few ways to obtain these artifacts:
```

README.md
```diff
@@ -1,58 +1,155 @@
...
-     ```
+- Have the container put the artifacts into a mounted volume.
+- Manually copy the artifacts to a mounted volume.
+- Extracting the artifacts from the docker container with a docker command.
 
-Alternatively, the build can be started directly by passing `build` to the
-`docker run` command:
+### Having the container build files into a mounted volume
 
+Since the files are not uploaded, they need to be retrieved out of the repo before the container exits.
```
bdunne, line 120:
```suggestion
Since the files are not uploaded, they need to be retrieved from the repo before the container exits.
```

bdunne: IIRC, all of the files still exist on the mounted directory after the container exits, correct?

kbrock: once the primary process exists, you can not access the container anymore.

kbrock:
ok:

- made all doc changes.
- introduced dockerignore with Gemfile.lock
- outstanding: how can you get files out of a container that has exited the primary process?

Think we are done and ready to go. Pretty sure point 3 above is not possible

Fryguy:
> outstanding: how can you get files out of a container that has exited the primary process?

1. don't use `--rm` (which causes the stopped container to be removed)
2. Use `docker cp` to copy the files
3. `docker rm` to remove the stopped container

README.md
kbrock: I think I'm wrong here. I'll revisit and fix.

Since I am wrong, this greatly simplifies things. (thanks to jason't comments down below)

kbrock:
ok, pushed again, removing a bunch of complexity I had added

mostly surrounding copying assets out of a container. `--rm` was the culprit for many of my difficulties

Fryguy (approved):
LGTM - @bdunne is this good?

bdunne (approved):

Fryguy:
Backported to `morphy` in commit 492becd490fa0ebe00757dfa4b51fdf2e3ae5171.

```text
commit 492becd490fa0ebe00757dfa4b51fdf2e3ae5171
Author: Brandon Dunne <bdunne@redhat.com>
Date:   Wed Jun 16 15:41:09 2021 -0400

    Merge pull request #152 from kbrock/document_build
    
    add common usage examples
    
    (cherry picked from commit f9908ca9eb9827729963a8a00c9d2e619066f29e)
```

PR #156, opened 2021-06-07
Add dnf update to get updated packages for s390x arch

dpkulkarni:
Added dnf update to get the updated packages on s390x arch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ARG ARCH=x86_64
 ENV TERM=xterm \
     APPLIANCE=true \
     RAILS_USE_MEMORY_STORE=true
+RUN dnf -y update
 
 RUN curl -L https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo > /etc/yum.repos.d/ansible-runner.repo
 


Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG ARCH=x86_64
 ENV TERM=xterm \
     APPLIANCE=true \
     RAILS_USE_MEMORY_STORE=true
+RUN dnf -y update
```
bdunne, line 8: Would moving to `ubi:8.4` also fix this issue?

dpkulkarni: Just updated the changes with `ubi:8.4` which also needs the dnf update.

Fryguy:
I don't really understand this change as it implies the base image doesn't have the right packages which doesn't make sense.  This image is already on ubi 8.4.

I'd rather _not_ have this change as dnf updates in containers incur a penalty in file sizes.

Fryguy:
Backported to `morphy` in commit dca2a8d3249cf351f2c2aec04735ba1128ae4f02.

```text
commit dca2a8d3249cf351f2c2aec04735ba1128ae4f02
Author: Brandon Dunne <bdunne@redhat.com>
Date:   Mon Jun 7 13:17:22 2021 -0400

    Merge pull request #156 from dpkulkarni/z_changes
    
    Add dnf update to get updated packages for s390x arch
    
    (cherry picked from commit 507aa27e1d30f205da6a8c06dae023f8059a3007)
```

PR #166, opened 2021-07-08
Remove . from RPM release number

bdunne:
- Change 0.1 to 0 in the config
- Switch from 0.1.timestamp to timestamp only in the release
- Add versioning information to the README (with some changes from the previous versioning)

We've been seeing issues with the appliance build where all of the ManageIQ RPMs are installed, but nightlies don't have the latest version of the RPM. After some reading, my best guess is that we're incorrectly using the release since we should follow `name-epoch:version-release.architecture` where `epoch:` is optional.

Nightly:
  Before: manageiq-core-13.0.0-0.1.20210708000051.el8.x86_64
  After:  manageiq-core-13.0.0-20210708000051.el8.x86_64
Pre-release:
  Before: manageiq-core-13.0.0-0.1.beta1.el8.x86_64
  After:  manageiq-core-13.0.1-beta1.el8.x86_64
Release:
  Before: manageiq-core-13.0.0-0.1.el8.x86_64
  After:  manageiq-core-13.1.0-1.el8.x86_64

Additional reading:
https://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,21 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
 docker rm ${CONTAINER}
 ```
 
+## Versioning
+
+Branch `morphy` == v13
+
+| Version               | Purpose     | Repo    |
+|-----------------------|-------------|---------|
+| 13.0.0-20210708000051 | nightly     | nightly |
+| 13.0.1-beta1          | pre-release | release |
+| 13.0.1-20210708000051 | nightly     | nightly |
+| 13.0.2-rc1            | pre-release | release |
+| 13.1.0-0              | release     | release |
+| 13.1.0-20210708000051 | nightly     | nightly |
+| 13.2.0-0              | release     | release |
+| 13.2.0-20210708000051 | nightly     | nightly |
+
 ## License
 
 This code is available as open source under the terms of the [Apache License 2.0](LICENSE).
--- a/config/options.yml
+++ b/config/options.yml
@@ -15,7 +15,7 @@ version:
 release:
 rpm:
   version:         13.0.0
-  release:         0.1
+  release:         0
   repo_name:
   org_name:        manageiq
   product_summary: ManageIQ Management Engine
--- a/lib/manageiq/rpm_build/build_copr.rb
+++ b/lib/manageiq/rpm_build/build_copr.rb
@@ -65,10 +65,10 @@ module ManageIQ
 
       def spec_release
         if release_name.empty?
-          "#{rpm_release}.#{BUILD_DATE}"
+          BUILD_DATE
         else
           pre_build = release_name.split("-")[2]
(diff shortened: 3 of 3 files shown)


bdunne:
@simaishi If you have a few minutes I'd appreciate your review on this too.

simaishi:
@bdunne `0.1` is just the initial one before the first pre-release and it should be updated to `0.2`, `0.3` as you do more pre-releases (to make version comparison works properly).

What is the version you expected to see installed and what is the version actually installed?

bdunne:
> @bdunne `0.1` is just the initial one before the first pre-release and it should be updated to `0.2`, `0.3` as you do more pre-releases (to make sure version comparison works properly). [#4 (comment)](https://github.com/ManageIQ/manageiq-rpm_build/issues/4#issuecomment-618633438) explains how the comparison works.
> 
> What is the version you expected to see installed and what is the version actually installed?

It's strange, we're getting a bunch of different versions. https://github.com/ManageIQ/manageiq-appliance-build/pull/483#issuecomment-876514420 All RPMs were successfully built and uploaded and have been between these dates as well.  Booting the image and running `yum -y update` pulls in the latest versions, so I'm thinking there might be a bug in anaconda, but I don't think we need the extra digits either.

simaishi:
As long as you increment something in each pre-release/release, it will work. I think I used the relaese (not version) to increment, because we sometimes had an extra release (e.g. hammer_5.1)

bdunne:
Removing backport labels since master appliances are still upgrading RPMs in the post script.  More investigation is necessary.

Fryguy:
@bdunne A conflict occurred during the backport of this pull request to `morphy`.

If this pull request is based on another pull request that has not been marked for backport, add the appropriate labels to the other pull request. Otherwise, please create a new pull request direct to the `morphy` branch in order to resolve this.

Conflict details:

```diff
diff --cc config/options.yml
index a5986a7,28bec96..0000000
--- a/config/options.yml
+++ b/config/options.yml
@@@ -14,14 -14,14 +14,19 @@@ repos
  version:
  release:
  rpm:
++<<<<<<< HEAD
 +  version: 13.0.0
 +  release: 0
++=======
+   version:         13.0.0
+   release:         0
++>>>>>>> 433a2df... Merge pull request #166 from bdunne/fix_release_number
    repo_name:
 -  org_name:        manageiq
 +  org_name: manageiq
    product_summary: ManageIQ Management Engine
 -  product_url:     https://github.com/ManageIQ/manageiq
 +  product_url: https://github.com/ManageIQ/manageiq
    changelog:
 -bundler_version:   2.1.4
 +bundler_version: 2.1.4
  npm_registry:
  rpm_repository:
    digitalocean_access_token:
```

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-rpm_build/pull/191

PR #173, opened 2021-07-22
Return descendent loader cache to regular privs

kbrock:
related to https://github.com/ManageIQ/manageiq/issues/20394

we were trying to overwrite the descenent loader cache
This was fixed in d4ecc500c8c ( https://github.com/ManageIQ/manageiq/pull/21321 )

now we no longer need the special/explicit privileges

### DIFF
--- a/rpm_spec/subpackages/manageiq-core
+++ b/rpm_spec/subpackages/manageiq-core
@@ -53,9 +53,6 @@ done
 %attr(-,manageiq,manageiq) %{app_root}/config
 %attr(-,manageiq,manageiq) %{app_root}/log
 %attr(-,manageiq,manageiq) %{app_root}/tmp
-## this is until we fix the code that auto writes this
-## give kbrock a hard time if it is 2022 and this is still in here
-%attr(0644,manageiq,manageiq) %{app_root}/tmp/cache/sti_loader.yml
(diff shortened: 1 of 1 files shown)


kbrock:
UNWIP:

installed this on an appliance and verified all are running.

Of note:
This really didn't do much. The permissions are the same before and after.
`tmp/` and all files in it are owned by 'manageiq.manageiq` so setting the ownership on this file in particular to `manageiq.manageiq` was a no-op.

I had tried to figure out if it made sense to have it owned by `root` but since this is a cached file, I think it does make sense to have it owned by `manageiq` even though we will never write to the file in production. This seems to go with @Fryguy reasoning around the file from what I could tell.

rpm_spec/subpackages/manageiq-core
```diff
@@ -53,9 +53,6 @@ done
 %attr(-,manageiq,manageiq) %{app_root}/config
 %attr(-,manageiq,manageiq) %{app_root}/log
 %attr(-,manageiq,manageiq) %{app_root}/tmp
-## this is until we fix the code that auto writes this
-## give kbrock a hard time if it is 2022 and this is still in here
```
agrare, line 57: With time to spare!

agrare (approved):

PR #177, opened 2021-08-19
Reduce disk space of locale files

kbrock:
requires:

- [x] https://github.com/ManageIQ/manageiq/pull/21406

~~there are comments at the top of the locale files~~
~~removing these comments reduces the disk space by half (15meg)~~

Using `.mo` files instead of `.po` files to reduce disk space.

testing w/ upgrading rpms on an existing appliance.

### Before

```
[root@localhost vmdb]# wc locale/*/*.po
   82152  246744 3068312 locale/de/manageiq.po
   83011  201527 2496587 locale/en/manageiq.po
   82152  274848 3127414 locale/es/manageiq.po
   82152  276191 3131189 locale/fr/manageiq.po
   82152  263833 3057572 locale/it/manageiq.po
   81604  216590 3167343 locale/ja/manageiq.po
   81604  258917 3090377 locale/ko/manageiq.po
   82152  272322 3142088 locale/pt_BR/manageiq.po
   81604  215609 2902947 locale/zh_CN/manageiq.po
   81604  217728 2917629 locale/zh_TW/manageiq.po
  820187 2444309 30,101,458 total
[root@localhost vmdb]# du -sh locale
 31M	locale
```

### After

```
[root@localhost vmdb]# wc locale/*/*.mo
    3581   113159  1493525 locale/de/manageiq.mo
      11       32      462 locale/en/manageiq.mo
    3480   140302  1552600 locale/es/manageiq.mo
    3491   141413  1556361 locale/fr/manageiq.mo
    3550   129736  1482781 locale/it/manageiq.mo
    3422    79023  1599177 locale/ja/manageiq.mo
    3472   111029  1522165 locale/ko/manageiq.mo
    3406   137673  1567245 locale/pt_BR/manageiq.mo
    3585    76814  1335809 locale/zh_CN/manageiq.mo
    3558    78617  1349450 locale/zh_TW/manageiq.mo
   31556  1007798 13,459,575 total
[root@localhost vmdb]# du -sh locale
16M	locale
```

## old

**No longer valid***, but keeping around for posterity sake.

### After

```
[root@localhost vmdb]# wc locale/*/*.po
   52907  170794 1454711 locale/de/manageiq.po
   53766  125577  882986 locale/en/manageiq.po
   52907  198898 1513813 locale/es/manageiq.po
   52907  200241 1517588 locale/fr/manageiq.po
   52907  187883 1443971 locale/it/manageiq.po
   52359  140640 1553742 locale/ja/manageiq.po
   52359  182967 1476776 locale/ko/manageiq.po
   52907  196372 1528487 locale/pt_BR/manageiq.po
   52359  139659 1289346 locale/zh_CN/manageiq.po
   52359  141778 1304028 locale/zh_TW/manageiq.po
  527737 1684809 13,965,448 total
[root@localhost vmdb]# du -sh locale
16M	locale
```

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -82,7 +82,7 @@ module ManageIQ
         precompile_sti_loader
         build_service_ui
         seed_ansible_runner
-        cleanse_locale_files
+        compile_locale_files
 
         if OPTIONS.npm_registry
           Dir.chdir(miq_dir) do
@@ -150,10 +150,14 @@ module ManageIQ
         end
       end
 
-      def cleanse_locale_files
-        # remove all comment lines from locale files
-        # the comment lines beginning with #, are special
-        shell_cmd("sed -i'' -e '/^#[^,]/d' #{miq_dir.join('locale/*/*.po')}")
+      def compile_locale_files
+        miq_dir.glob("locale/*/*.po").each do |po|
+          mo = po.dirname.join("LC_MESSAGES", "#{File.basename(po, '.po')}.mo")
+          mo.dirname.mkpath
+          shell_cmd("msgfmt #{po} -o #{mo}")
+        end
+        shell_cmd("rm #{miq_dir.join('locale/*/*.po')}")
+        shell_cmd("rm #{miq_dir.join('locale/*.pot')}")
       end
     end
   end
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -82,6 +82,7 @@ module ManageIQ
         precompile_sti_loader
         build_service_ui
         seed_ansible_runner
+        cleanse_locale_files
 
         if OPTIONS.npm_registry
           Dir.chdir(miq_dir) do
@@ -147,6 +148,12 @@ module ManageIQ
             FileUtils.ln_s(path.join(subdir), target_path.join(subdir))
           end
         end
+      end
+
+      def cleanse_locale_files
+        # remove all comment lines from locale files
+        # the comment lines beginning with #, are special
+        shell_cmd("sed -i'' -e '/^#[^,]/d' #{miq_dir.join('locale/*/*.po')}")
       end
     end
   end


Fryguy:
Nice!  Would also like @chessbyte to review, size he's the in the i18n space.

chessbyte:
From the [PO File Format](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html), "The comment lines beginning with #, are special because they are not completely ignored by the programs as comments generally are. The comma separated list of flags is used by the msgfmt program to give the user some better diagnostic messages."

So, I think we can delete all comments except those starting with `#,`

Fryguy:
Oh great eye @chessbyte  - I don know we do use at least one flag (around line 5 of most .po files):

```
#, fuzzy
```

kbrock:
Great find.
Looks like we are only keeping 11 extra characters (110bytes in total)
```
#
#, fuzzy
```
Making a more complicated regex to remove the blank comment line (2 characters) is not worth the risk for me.

Savings is still 15.4mb

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -148,6 +149,12 @@ def symlink_plugin_paths(source_dir, target_path)
           end
         end
       end
+
+      def cleanse_locale_files
+        # remove all comment lines from locale files
+        # the comment lines beginning with #, are special
+        `sed -i '' -e '/^#[^,]/d' #{miq_dir.join('locale/*/*.po')}`
```
Fryguy, line 156:
```suggestion
        shell_cmd("sed -i '' -e '/^#[^,]/d' #{miq_dir.join('locale/*/*.po')}")
```

Fryguy: Also we have to be careful here with sed.  I think the -i works differently on Mac vs Linux.  @kbrock where did you run this command?

Fryguy:
While I like this (and will ultimately merge), separtely I wonder if we should consider generating .mo files at runtime.  Can fast-gettext support that?  Maybe @jrafanie knows offhand.

Locally I found:

```sh
$ for f in $(find . -name *.po); do; msgfmt $f -o ${f%.po}.mo; done
$ diff -y <(ls -al **/*.po | awk '{print $9,$5}') <(ls -al **/*.mo | awk '{print $9,$5}')
locale/de/manageiq.po 3068312				      |	locale/de/manageiq.mo 1493484
locale/en/manageiq.po 2496587				      |	locale/en/manageiq.mo 421
locale/es/manageiq.po 3127414				      |	locale/es/manageiq.mo 1552559
locale/fr/manageiq.po 3131189				      |	locale/fr/manageiq.mo 1556320
locale/it/manageiq.po 3057572				      |	locale/it/manageiq.mo 1482740
locale/ja/manageiq.po 3167343				      |	locale/ja/manageiq.mo 1599136
locale/ko/manageiq.po 3090377				      |	locale/ko/manageiq.mo 1522124
locale/pt_BR/manageiq.po 3142088			      |	locale/pt_BR/manageiq.mo 1567204
locale/zh_CN/manageiq.po 2902947			      |	locale/zh_CN/manageiq.mo 1335768
locale/zh_TW/manageiq.po 2917629			      |	locale/zh_TW/manageiq.mo 1349409
public/ui/service/gettext/po/de/manageiq-ui-service.po 59117  |	public/ui/service/gettext/po/de/manageiq-ui-service.mo 25167
public/ui/service/gettext/po/es/manageiq-ui-service.po 58795  |	public/ui/service/gettext/po/es/manageiq-ui-service.mo 24845
public/ui/service/gettext/po/fr/manageiq-ui-service.po 59251  |	public/ui/service/gettext/po/fr/manageiq-ui-service.mo 25301
public/ui/service/gettext/po/it/manageiq-ui-service.po 58445  |	public/ui/service/gettext/po/it/manageiq-ui-service.mo 24495
public/ui/service/gettext/po/ja/manageiq-ui-service.po 60274  |	public/ui/service/gettext/po/ja/manageiq-ui-service.mo 26335
public/ui/service/gettext/po/ko/manageiq-ui-service.po 58764  |	public/ui/service/gettext/po/ko/manageiq-ui-service.mo 24826
public/ui/service/gettext/po/nl/manageiq-ui-service.po 55096  |	public/ui/service/gettext/po/nl/manageiq-ui-service.mo 21995
public/ui/service/gettext/po/pt-BR/manageiq-ui-service.po 584 |	public/ui/service/gettext/po/pt-BR/manageiq-ui-service.mo 244
public/ui/service/gettext/po/pt-PT/manageiq-ui-service.po 484 |	public/ui/service/gettext/po/pt-PT/manageiq-ui-service.mo 365
public/ui/service/gettext/po/zh-CN/manageiq-ui-service.po 556 |	public/ui/service/gettext/po/zh-CN/manageiq-ui-service.mo 217
public/ui/service/gettext/po/zh-TW/manageiq-ui-service.po 562 |	public/ui/service/gettext/po/zh-TW/manageiq-ui-service.mo 223
```

This is roughly the same savings that @kbrock found, and it might also give us system savings if parsing mo files is more efficient.

lib/manageiq/rpm_build/generate_core.rb
kbrock: ran on mac and it worked fine
ran this through the full rpm build process (docker linux) and it ran fine.

the `-i ''` is what I do to make it work better with both.
Pretty sure on linux you do not need the `''`

Fryguy: I'm concerned that on linux it will blow up - maybe @bdunne or @agrare can try?

agrare: `sed -i ''` doesn't seem to work:
```bash
adam@desktop:~/src/manageiq/manageiq$ sed -i '' -e '/^#[^,]/d' locale/en/manageiq.po 
sed: can't read : No such file or directory
adam@desktop:~/src/manageiq/manageiq$ sed -i -e '/^#[^,]/d' locale/*/*.po
```

The `sed(1)` man page says:
>        -i[SUFFIX], --in-place[=SUFFIX]
>            edit files in place (makes backup if SUFFIX supplied)

So I think what you want is actually:
```bash
adam@desktop:~/src/manageiq/manageiq$ sed -i'' -e '/^#[^,]/d' locale/*/*.po 
```

Aka the `''` immediately after the `-i` with no space.

kbrock:
@Fryguy ruby `GetText` only supports `mo` files while `FastGetText` supports `mo`, `po` and others. [ref]

1. Do we actually gain something by going over to `mo` files?
2. Should this change be part of the localization/git checkin process instead of the build process?
3. Do we want to instead add json approach to `FastGetText` (assuming the json format is standardized in the js world)? It is upgrading an existing dependency but grosser is very active and responsive.

[ref]: https://github.com/grosser/fast_gettext#comparison

jrafanie:
> 1\. Do we actually gain something by going over to `mo` files?

@Fryguy is suggesting that `mo` files generated at build time are much smaller than `po`.  So, in development, we can use `po` files which are supposed to be easier to maintain, but then generate smaller `mo` files at build time to be used at runtime.  


> 2\. Should this change be part of the localization/git checkin process instead of the build process?

I don't know how much work maintaining `po` vs `mo` is in development. The recommendation is to use traditional `mo` in the build process to produce smaller production payloads/rpms/etc.  If we need `mo` in development, that's a whole other story that's worth reviewing with whoever does translations.


> 3\. Do we want to instead add json approach to `FastGetText` (assuming the json format is standardized in the js world)? It is upgrading an existing dependency but grosser is very active and responsive.

Haha, queue the [14 standards](https://xkcd.com/927/).  I noticed they, fast gettext, already also supports yaml and database storage.  I don't know what's involved in supporting json or what it would gain us but it seems like there's a huge gain in just using the existing `mo` format.  So, depends if we want to investigate it and if that time is better served using `mo` and work on other issues.  I don't know.

kbrock:
WIP: added `mo` conversions. testing a build now.

if this is WIP 8/26/21 then I flaked and need to un-WIP this.

lib/manageiq/rpm_build/generate_core.rb
kbrock: no more `sed`

kbrock:
un-WIP:

- [x] no `locale` error in error pages
- [x] localized content on pages when you change chrome default language
- [x] `locale/*/manageiq.mo` files
- [x] no `locale/*/manageiq.po` files
- [ ] no errors on rpm install

kbrock:
@Fryguy this what you had in mind?

kbrock:
UNWIP: fixed `mo` files - they are not being picked up by the rails.

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -148,6 +149,15 @@ def symlink_plugin_paths(source_dir, target_path)
           end
         end
       end
+
+      def cleanse_locale_files
```
Fryguy, line 153: Since this is not a straight delete, but actually a proper step in the build process, let's rename this method... Something like `def compile_locales`

Fryguy (approved):
This LGTM - I think it can be a little cleaner with Pathnames, but overall I'm good with this.  I want @chessbyte to double check though.

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -148,6 +149,15 @@ def symlink_plugin_paths(source_dir, target_path)
...
       end
+
+      def compile_locale_files
+        Dir[miq_dir.join('locale/*/*.po')].each do |po|
+          mo = "#{File.dirname(po)}/LC_MESSAGES/#{File.basename(po, '.po')}.mo"
+          shell_cmd("mkdir #{File.dirname(mo)} ; msgfmt #{po} -o #{mo}")
+        end
+        shell_cmd("rm #{miq_dir.join('locale/*/*.po')}")
+        shell_cmd("rm #{miq_dir.join('locale/*.pot')}")
+      end
```
Fryguy, line 160: Think this might be a little cleaner, but I understand that testing _again_ is a pain, so I'm fine with it as is, if you want...

```suggestion
      def compile_locale_files
        miq_dir.glob('locale/*/*.po').each do |po|
          mo = po.dirname.join('LC_MESSAGES', "#{po.basename('.po')}.mo")
          mo.dirname.mkpath
          shell_cmd("msgfmt #{po} -o #{mo}")
        end
        shell_cmd("rm #{miq_dir.join('locale/*/*.po')}")
        shell_cmd("rm #{miq_dir.join('locale/*.pot')}")
      end
```

chessbyte (approved):
LGTM

lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -148,6 +149,15 @@ def symlink_plugin_paths(source_dir, target_path)
           end
         end
       end
+
+      def compile_locale_files
+        Dir[miq_dir.join('locale/*/*.po')].each do |po|
+          mo = "#{File.dirname(po)}/LC_MESSAGES/#{File.basename(po, '.po')}.mo"
+          shell_cmd("mkdir #{File.dirname(mo)} ; msgfmt #{po} -o #{mo}")
```
jrafanie, line 156: I guess `msgfmt` comes from `gettext`.  Do we need to add that to the list of packages in the Dockerfile?

https://github.com/ManageIQ/manageiq-rpm_build/blob/10200c9129542cf0585da2b210cc189bcdb2601b/Dockerfile#L14-L46

Fryguy: Oh yeah good catch

kbrock: Huh. wonder why it is working though.

lib/manageiq/rpm_build/generate_core.rb
kbrock: I'll update and test again.

lib/manageiq/rpm_build/generate_core.rb
jrafanie: I don't know if something else is enabling `msgfmt`.  Maybe try to see if you can get it to fail and complain about a missing `msgfmt` first, then we can try to enable whatever module that provides it.  It should fail right now unless something else is making it available as a direct or indirect dependency

kbrock: yea, I had already tested this a number of times (building the rpms / upgrading the appliance. not a brand new appliance)

```
[root@localhost ~]# rpm -qf /usr/bin/msgfmt
gettext-0.19.8.1-17.el8.x86_64
[root@localhost ~]# rpm -q --whatrequires gettext
grub2-tools-minimal-2.02-99.el8.x86_64
grub2-tools-2.02-99.el8.x86_64
grub2-tools-extra-2.02-99.el8.x86_64
grub2-pc-2.02-99.el8.x86_64
```

Looks like `grub2` sepends upon `gettext` -- so this is not going anywhere.

jrafanie: So, to answer my quesion:

> I guess `msgfmt` comes from `gettext`. Do we need to add that to the list of packages in the Dockerfile?

No, w can't get very far without grub2 and grub2 depends on gettext which provides the command we're trying to run so we don't have to explicitly require the gettext rpm.

kbrock:
update: made jasons changes
update2: fixed rubocop

This should be good to go

PR #180, opened 2021-08-27
Don't install gitkeep files

kbrock:
No reason to put `.gitkeep` files into our rpm.
Those files are a mechanism to tell `git` to create a directory.
Directories are just a side effect for `git` so we create files to force the creation of a directory.


Fryguy:
Will the directories still be in place?  I'm concerned this change will kill the directories.

I was going to suggest that if we remove these we should do so at the tarball generation stage? (Anytime I see things that mess with the filesystem _after_ the tarball, I always question why not just doing beforehand in the tarball)  However, if we remove it from the tarball, does that kill the empty dirs?

kbrock:
> Will the directories still be in place? I'm concerned this change will kill the directories.

the `.gitkeep` is there to tell `git` that these directories exist.

`tar` seem to respect empty directories:

```
manageiq $ mkdir a
manageiq $ mkdir a/b
manageiq $ tar -cvf a.tar a
a a
a a/b
manageiq $ rm -rf a
manageiq $ tar -xzf a.tar
manageiq $ ls a
b
```

`rpm` has a similar but different different issue around directories. `attr` in `%filelist` only applies to files not directories. This is why I apply `chmod` to directories and not in the `%filelist` section. the `dir` directive is strange and typically used wrong.

**WIP:** I will update the rpms and ensure that the directories are getting created.

> 
> I was going to suggest that if we remove these we should do so at the tarball generation stage? (Anytime I see things that mess with the filesystem _after_ the tarball, I always question why not just doing beforehand in the tarball) However, if we remove it from the tarball, does that kill the empty dirs?

The exact timing of the tarball confuses me. I will look into this.
But from what I know about `tar`, I do not see this as a problem.

**WIP:** I will move this to the tar creation portion.

kbrock:
unWIP: the directories are created and this work is moved to the tar

```
[root@localhost vmdb]# \rm public/pictures/.gitkeep
[root@localhost vmdb]# rmdir public/pictures
[root@localhost vmdb]# yum update -y /opt/rpms/*.rpm
...
Complete!
[root@localhost vmdb]# ls -la public/pictures
total 0
drwxr-xr-x. 2 manageiq manageiq   6 Sep  2  2021 .
drwxr-xr-x. 8 root     root     176 Sep  1 19:20 ..
[root@localhost vmdb]# find . -name .gitkeep
[root@localhost vmdb]#
```

Fryguy:
wow I did not expect that to be so tiny haha.  I think sometimes we also use .keep files, so should that be excluded too?

config/exclude_manageiq
```diff
@@ -10,3 +10,4 @@ Gemfile.lock*
 ./bin
 certs/v2_key.dev
 log/*
+.gitkeep
```
Fryguy, line 13: Also can you sort?

kbrock:
update: moved g above `L` but below `c` -- not exactly sure how to resort this file. 

Moving `Gemfile.lock*` down is the only other change I could see but nothing looked right. (sort sounds like such a simple request)

Fryguy:
Think this comment got lost:

>  I think sometimes we also use .keep files, so should that be excluded too?

For example: https://github.com/ManageIQ/manageiq-api/blob/master/bundler.d/.keep
There are a lot if you do a search.

kbrock:
> Think this comment got lost:
> 
> > I think sometimes we also use .keep files, so should that be excluded too?
> 
> For example: https://github.com/ManageIQ/manageiq-api/blob/master/bundler.d/.keep
> There are a lot if you do a search.

so we want to search for `.gitkeep` and `.keep` files?
Darn. I thought `.gitkeep` was the convention.

kbrock:
> This PR got me thinking - why do we cleanse the gemset in Ruby by deleting files, but we cleanse manageiq by using an exclude file? It feels like we should pick one or the other, unless it's a "complicated" exclusion. Thoughts? (Not for this PR - I'm going to merge anyway, just wanted to spark the discussion)

if we remove the `rm` statements corrupting our gems, then we probably can reuse a container to build multiple times. this would speed things up for me significantly

PR #183, opened 2021-09-02
Add travis script to build the rpm_build container image

bdunne:


.travis.yml
```diff
@@ -0,0 +1,4 @@
+jobs:
+  include:
+    - stage: build and push container image
+      script: bin/build_container_image
```
Fryguy, line 4: I'm good with this for now, but ideally this should be split into 2 separate parts.

The primary job should just be to build and tag.  This can (must?) build on PRs.
The second half should be a deploy step, and that should only run on master.

bin/build_container_image
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+REGISTRY=${REGISTRY:-"registry.gitlab.com"}
```
Fryguy, line 3: I thought we were still going to use the docker registry (at least until they kill it).  I think we have too much pointing to docker.io right now that I don't think we should change it.

.travis.yml
bdunne: - The build and push need to happen in the same "stage" since different stages get their own VM, so we wouldn't have access to anything built in the first stage.  That's why I put that logic in the script rather than in the `.travis.yml`.
- I thought it would still be a good idea to only build on PRs so that we catch any Dockerfile errors before it's merged.

bin/build_container_image
bdunne: I can't add any users to our docker.io org, we're already using 5 of 3 seats.  So, I thought it would be good to try out the Gitlab registry with this image before moving everything else over.

.travis.yml
Fryguy: Ah that's frustrating

:+1:

bin/build_container_image
Fryguy: I can adjust the users - that's the easy part.  That being said, the bot is already a user and should be the one we use (it's the same one we use for the build).

.travis.yml
bdunne: Switched to docker.io with the bot user and a token

bin/build_container_image
bdunne: Switched to docker.io with the bot user and a token

Fryguy:
Backported to `morphy` in commit c8a8c622ca1bc0d8ae9103e69d2d2188326083eb.

```text
commit c8a8c622ca1bc0d8ae9103e69d2d2188326083eb
Date:   Tue Sep 7 10:02:37 2021 -0400

    Merge pull request #183 from bdunne/build_container_image
    
    Add travis script to build the rpm_build container image
    
    (cherry picked from commit 10200c9129542cf0585da2b210cc189bcdb2601b)
```

PR #184, opened 2021-09-07
Travis build script updates

bdunne:
Followup to https://github.com/ManageIQ/manageiq-rpm_build/pull/183
- Set tag dynamically based on branch name
- Prefer to send the password to docker login through stdin

### DIFF
--- a/bin/build_container_image
+++ b/bin/build_container_image
@@ -2,7 +2,9 @@
 
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
-TAG=${TAG:-"latest"}
+DEFAULT_TAG="latest"
+[[ $TRAVIS_BRANCH != "master" ]] && DEFAULT_TAG="$DEFAULT_TAG-$TRAVIS_BRANCH"
+TAG=${TAG:-"$DEFAULT_TAG"}
 IMAGE_NAME=$REGISTRY/$ORGANIZATION/rpm_build:$TAG
 
 set -e
--- a/bin/build_container_image
+++ b/bin/build_container_image
@@ -11,7 +11,7 @@ docker build . -t $IMAGE_NAME
 
 [[ $TRAVIS_PULL_REQUEST != "false" ]] && exit 0
 
-docker login $REGISTRY -u $REGISTRY_USERNAME -p $REGISTRY_PASSWORD
+echo "$REGISTRY_PASSWORD" | docker login $REGISTRY -u $REGISTRY_USERNAME --password-stdin
 
 docker push $IMAGE_NAME
 


bin/build_container_image
```diff
@@ -2,7 +2,8 @@
 
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
-TAG=${TAG:-"latest"}
+[[ $TRAVIS_BRANCH == "master" ]] && TAG=latest
```
Fryguy, line 5: We still want to use the TAG if it was passed in, right? If so, then I think this should be

```suggestion
[[ $TRAVIS_BRANCH == "master" ]] && TAG=${TAG:-latest}
```

Or perhaps it's better to be more explicit?  Something like:

```sh
DEFAULT_TAG="latest"
[[ $TRAVIS_BRANCH != "master" ]] && DEFAULT_TAG="$DEFAULT_TAG-$TRAVIS_BRANCH"
TAG=${TAG:-"$DEFAULT_TAG"}
```

bin/build_container_image
```diff
@@ -11,7 +12,7 @@ docker build . -t $IMAGE_NAME
 
 [[ $TRAVIS_PULL_REQUEST != "false" ]] && exit 0
 
-docker login $REGISTRY -u $REGISTRY_USERNAME -p $REGISTRY_PASSWORD
+echo $REGISTRY_PASSWORD | docker login $REGISTRY -u $REGISTRY_USERNAME --password-stdin
```
Fryguy, line 15: Minor, but we should quote this just in case it has some special characters in it.

```suggestion
echo "$REGISTRY_PASSWORD" | docker login $REGISTRY -u $REGISTRY_USERNAME --password-stdin
```

bin/build_container_image
```diff
@@ -2,7 +2,8 @@
 
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
-TAG=${TAG:-"latest"}
+[[ $TRAVIS_BRANCH == "master" ]] && TAG=latest
+TAG=${TAG:-"latest-${TRAVIS_BRANCH}"}
```
Fryguy, line 6: Will this cause problems if run locally?  Do we have to check for the presence of $TRAVIS_BRANCH first?  I'm not sure how a local run is expected to work and what we should default to (I assume latest?)

Fryguy (approved):

bin/build_container_image
bdunne: Done

Fryguy:
Backported to `morphy` in commit 0016f6bc1db20d2f375a5ecf905351e09478d90d.

```text
commit 0016f6bc1db20d2f375a5ecf905351e09478d90d
Date:   Tue Sep 7 15:56:33 2021 -0400

    Merge pull request #184 from bdunne/travis
    
    Travis build script updates
    
    (cherry picked from commit a91fe405b913f45eb9911f9c7c1f181329fee1db)
```

PR #185, opened 2021-09-07
Replace rake update_{stop,start} with systemctl try-restart evmserverd

agrare:
Instead of stopping evmserverd, writing no_db to the pid file, then starting it again after simply issue try-restart after rpms finish install

Part of: https://github.com/ManageIQ/manageiq-appliance/pull/327

### DIFF
--- a/rpm_spec/subpackages/manageiq-appliance
+++ b/rpm_spec/subpackages/manageiq-appliance
@@ -55,29 +55,8 @@ Requires: yum-utils
 %description appliance
 %{product_summary} Appliance
 
-%pretrans appliance -p <lua>
-if posix.access('/bin/evmserver.sh', 'x') then
-  local pid = posix.fork ()
-  if ( pid == -1 ) then
-    print ("The fork failed.")
-  elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_stop')
-  else
-    posix.wait(pid)
-  end
-end
-
-%posttrans appliance -p <lua>
-if posix.access('/bin/evmserver.sh', 'x') then
-  local pid = posix.fork ()
-  if ( pid == -1 ) then
-    print ("The fork failed.")
-  elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_start')
-  else
-    posix.wait(pid)
-  end
-end
+%posttrans appliance
+%{_bindir}/systemctl try-restart evmserverd.service
 
 %post appliance
 #motd is owned by system


rpm_spec/subpackages/manageiq-appliance
```diff
@@ -55,29 +55,8 @@ Requires: yum-utils
...
-  local pid = posix.fork ()
-  if ( pid == -1 ) then
-    print ("The fork failed.")
-  elseif ( pid == 0 ) then
-    posix.exec('/bin/evmserver.sh', 'update_start')
-  else
-    posix.wait(pid)
-  end
-end
+posix.exec("%{_bindir}/systemctl", "try-restart", "evmserverd.service")
```
Fryguy, line 59: ❤️  This is so much nicer!

Fryguy: Interestingly, the "old" code seems to follow a general fork/exec pattern, but this does a straight exec.  Is that right or should it fork/exec like the old way?

agrare: Yeah I was excited to learn about try-restart, thought I was going to have to do systemctl status, check the return then restart if it was already running

Fryguy: I think it has to fork/exec/wait like the old way though, still.  I recall that being an oddity either with Lua or perhaps just RPM itself - maybe @bdunne remembers.

Fryguy:
@bdunne Please review.

rpm_spec/subpackages/manageiq-appliance
agrare: > Interestingly, the "old" code seems to follow a general fork/exec pattern

Yeah possibly, need to look into if it is required

agrare: Do we even need lua here now?  Seeing some examples here https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd where they just call `systemctl restart package.service`

Fryguy: Probably not...if there's already macros for systemd directly, we should probably use those.

agrare: Testing this out now, will update back with what I find

agrare: Yeah we can simply run `systemctl try-restart evmserverd` without any other complexity

Fryguy (approved):

jrafanie (approved):
LGTM if we're not concerned with update_start anymore

PR #189, opened 2021-09-10
Fix URLs for Najdorf nightly

bdunne:


kbrock (approved):
pointing to the new base url looks great.

little confused about the release change, but sounds like that happens when the repos change

Fryguy:
This should have been handled by the release script.  If it's broken, can you also fix that?

PR #190, opened 2021-09-14
Ensure that we use the bundler version specified in options.yml

bdunne:
This was using bundler version 1.17.2 from the RPM for the build process.


Dockerfile
```diff
@@ -53,4 +52,6 @@ RUN echo "gem: --no-ri --no-rdoc --no-document" > /root/.gemrc
 
 COPY . /build_scripts
 
+RUN gem install bundler -v $(ruby -e "require 'yaml'; puts YAML.load_file('/build_scripts/config/options.yml')['bundler_version']")
```
Fryguy, line 55: Minor...

```suggestion
RUN gem install bundler -v $(ruby -e -r yaml "puts YAML.load_file('/build_scripts/config/options.yml')['bundler_version']")
```

Fryguy, line 55: Also this doesn't honor the overrides that can be placed in the OPTIONS directory.

kbrock:
I do notice that we install bundler the first time. Then we install it again from the manageiq core.
(not sure if you remember but I was running into problems when I locally had a bundler 1 and manageiq wanted bundler 2 - or something like that)

Would be cool if we could just run bundler once.
Not sure if it is possible to remove ruby from the process before downloading `manageiq`

PR #198, opened 2021-09-20
[MORPHY] Followup from #166

bdunne:
Followup from #166

(cherry picked from commit 81f1343efa936e7246d844ce30b2453033296e70)


Fryguy:
The "13" bits should have been handled by the release script, but let's do that in a follow up

PR #202, opened 2021-09-28
Remove postfix dependency since the application does not rely on it

bdunne:
Mail can be sent using ActionMailer without it

### DIFF
--- a/rpm_spec/subpackages/manageiq-system
+++ b/rpm_spec/subpackages/manageiq-system
@@ -4,9 +4,6 @@ Requires: hostname
 Requires: iproute
 Requires: nmap-ncat
 
-# For Email
-Requires: postfix
-
 # For log rotate
 Requires: cronie
 Requires: logrotate


Fryguy:
This was added in https://github.com/ManageIQ/manageiq-appliance-build/pull/349 - maybe @jrafanie can give more info.

Fryguy:
Spoke with @jrafanie and @chessbyte  and this is for hosting an internal mail server, but that doesn't seem very useful in a real production deployment, because a user would always want to use a proper SMTP server for sending mail.  External mail servers are not affected by this.  So, decision is to remove postfix.

Fryguy:
@bdunne The original code had a `systemctl enable postfix` as well...can you remove that too?

Fryguy:
Backported to `morphy` in commit 2fc2ad46001e8dddd8e9775028fc28aba01fd161.

```text
commit 2fc2ad46001e8dddd8e9775028fc28aba01fd161
Date:   Tue Sep 28 16:12:36 2021 -0400

    Merge pull request #202 from bdunne/remove_postfix
    
    Remove postfix dependency since the application does not rely on it
    
    (cherry picked from commit 7bf0eb238e1dc2b9b25658514952b8fde9a9679a)
```

PR #205, opened 2021-09-29
[WIP] install files directly to target location vs COPY/LINK

kbrock:
## Background

Dependency:

- [ ] https://github.com/ManageIQ/manageiq-appliance/pull/336

We used to have our appliances based upon a git repo.
So our workflow was focused on making this work as easily as possible.

An rpm workflow works a little differently.
And that workflow works better if we put the files at their final destination

## Before

- Install all files to one directory (`/opt/manageiq/manageiq-appliance/`).
- Copy files from `/opt/manageiq/manageiq-appliance/COPY` to the root directory (`/`).
- Link files from a `/opt/manageiq/manageiq-appliance/LINK` to the root directory (`/`).
- Copy user modifiable configuration files from a template to the final name (sometimes overwriting them)

## After

- Install all files to the root directory (`/`).
- Any user modified configuration files are saved as `rpmsave`

## Changes

```
rpm -ql manageiq-appliance # list all appliance specific files
rpm -V manageiq-appliance # list any changes we have made to the files
```

- There are close to half the number of files showing up in our list of files since we no longer are installing the files and a copy of them.
- Verify now shows when configuration files have changed.


kbrock:
Tested: I installed rpms on a 5/6/21 appliance and all seems to be working.

WIP reason: Would like a few reviews first.

Fryguy:
Are we going to have problems since we are converting symlinks to hard files on upgrade?  I recall an issue with this in the past - @bdunne do you remember?  @kbrock Did you try an upgrade?

Fryguy:
See https://github.com/ManageIQ/manageiq/pull/18683#issuecomment-496504383  (Thanks @simaishi !)

kbrock:
I did an rpm upgrade. (I am not able to do a clean appliance build)

It looks like everything worked. This is changing a symlink with a file.

```
[root@localhost vmdb]# ls -ld /usr/lib/systemd/system/manageiq*service
-rw-r--r--. 1 root root 445 Sep 29 17:35 /usr/lib/systemd/system/manageiq-cockpit_ws@.service
-rw-r--r--. 1 root root 316 Sep 29 17:35 /usr/lib/systemd/system/manageiq-db-ready.service
-rw-r--r--. 1 root root 456 Sep 29 17:35 /usr/lib/systemd/system/manageiq-ems_metrics_processor@.service
-rw-r--r--. 1 root root 419 Sep 29 17:35 /usr/lib/systemd/system/manageiq-event_handler@.service
-rw-r--r--. 1 root root 402 Sep 29 17:35 /usr/lib/systemd/system/manageiq-generic@.service
-rw-r--r--. 1 root root 351 Sep 29 17:35 /usr/lib/systemd/system/manageiq-initialize.service
-rw-r--r--. 1 root root 406 Sep 29 17:35 /usr/lib/systemd/system/manageiq-priority@.service
[...]
```

Fryguy:
ah ok...I think symlinked _files_ to real files works...it's just _directories_ that are a problem.  If there are no directories, then we are good.

PR #208, opened 2021-10-07
Drop ovirt-ansible-collection

agrare:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/570


Fryguy:
This will also kill python3-ovirt-engine-sdk4

PR #214, opened 2021-10-14
Revert "Move initial bundle into the Docker image"

Fryguy:
This reverts commit 30a904f4cab6cce853697b4e0cdc2c76db44e206.

This issue was introduced by #212 .  Reverting until we can solve it completely.

```
---> find "$GEM_HOME/specifications" -type l -xtype l -delete
/build_scripts/lib/manageiq/rpm_build/generate_core.rb:75:in `populate': Error: /root/BUILD/manageiq-gemset-14.0.0/bin/rake should be used, but /bin/rake is being used instead. (RuntimeError)
	from bin/build.rb:29:in `<main>'
```


Fryguy:
I think I figured out the actual problem, but the solution might be a little tricky...and it also takes forever and a day each iteration haha.

PR #217, opened 2021-10-20
Reduce image size by ~60MB

bdunne:
- Move the dnf update into the same layer that cleans dnf cache
- Disable subscription-manager before any dnf commands

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,13 +4,10 @@ ENV TERM=xterm \
     APPLIANCE=true \
     RAILS_USE_MEMORY_STORE=true
 
-RUN dnf -y update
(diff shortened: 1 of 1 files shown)


Fryguy:
Nice! I was gonna do this too, so glad you got it first.  Can you also add, after the dnf clean all:

```ruby
rm -rf /var/cache/dnf
```

We do this in other images too, like here: https://github.com/ManageIQ/manageiq-pods/blob/master/images/manageiq-base/Dockerfile#L59

bdunne:
> ```ruby
> rm -rf /var/cache/dnf
> ```

Done.

Issue #219, opened 2021-10-21
Master appliance build failure


bdunne:
Master appliance builds have been failing with `Error in POSTIN scriptlet in rpm package manageiq-core` see:
![factory-build-28da4639-7cf2-4982-827d-9ae3ba9dd226-1634657378 141397](https://user-images.githubusercontent.com/651659/138349097-53e9e8ab-f79d-4ff3-9672-00dd1de7670c.jpg)

kbrock:
So the rpms work for upgrade. They fail in Kickstarter. So there is something in here that is part of our appliance but is not part of the base layer install.

I said I would look into user creation (manageiq user). I need to circle back to see if we need to declare any other requirements for that

kbrock:
@bdunne are we all set to close this now?

bdunne:
Master builds are working, closing

PR #221, opened 2021-10-22
declare rpm dependencies for creating user

kbrock:
Attempting to address: https://github.com/ManageIQ/manageiq-rpm_build/issues/219

We create a manageiq user to properly setup
permissions as the manageiq (non-root) user

There were some hints from https://access.redhat.com/solutions/5749321
Missing dependencies seems to be a common `rpm` fail, especially when the dependencies are low level. The kickstarter file is probably the only case that something like this will fail.

These are basic dependencies, shadow is how we create all users and core-utils is well, basic stuff like `cp` and `ln` (one of the reasons I went with `id` instead of `getent` (glibc-common which is probably just as low level)

I have not built nor installed this. But as of recently I can no longer install self compiled manageiq rpms so it is beyond me to test.

### DIFF
--- a/rpm_spec/subpackages/manageiq-core
+++ b/rpm_spec/subpackages/manageiq-core
@@ -14,12 +14,14 @@ Requires: net-snmp
 Requires: net-snmp-libs
 Requires: net-snmp-utils
 Requires: socat
+Requires: core-utils
+Requires: shadow-utils
 
 %description core
 %{product_summary} Core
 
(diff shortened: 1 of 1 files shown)


kbrock:
I would still like to declare these dependencies

kbrock:
can't reopen this, so opened https://github.com/ManageIQ/manageiq-rpm_build/pull/406

PR #229, opened 2021-12-08
Fix RPM repo includes

bdunne:
- Add RSpec
- Add some initial tests for the RPM repo contents
- Run tests on Travis
- Fix regex to match current RPM versioning

PR #230, opened 2021-12-08
[MORPHY] Fix RPM repo includes

bdunne:
Manual cherry-pick from https://github.com/ManageIQ/manageiq-rpm_build/pull/229

(cherry picked from commit f7f70957f3f994d09bf3cd4098cf9eaf271b9564)


.travis.yml
```diff
@@ -1,6 +1,12 @@
 ---
+language: ruby
+cache: bundler
+rvm:
+- 2.5.8
+- 2.6.6
```
Fryguy, line 6: On morphy we only need 2.6

spec/config/options_spec.rb
```diff
@@ -0,0 +1,79 @@
+RSpec.describe "options.yml" do
+  let(:options) { ManageIQ::RPMBuild::OPTIONS }
+
+  nightly_versions = [
+    "13.1.0-20211208000053",
```
Fryguy, line 5: We are going to have to probably change the branching script to deal with these specs, since they are branch-specific, otherwise these are going to break next time we branch.

bdunne: I've got an idea, I'll make a master PR

bdunne: https://github.com/ManageIQ/manageiq-rpm_build/pull/231

PR #231, opened 2021-12-08
Pull the major version from the config file

bdunne:
This reduces the number of things that need to be updated when cutting a new branch


spec/config/options_spec.rb
```diff
@@ -57,8 +58,8 @@
     let(:repo) { "14-najdorf" }
```
Fryguy, line 58: Can we change this too?

bdunne: Done

PR #239, opened 2021-12-20
[WIP] Fix TRAVIS_BRANCH

bdunne:
- This appears to be blank on master

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,5 +32,5 @@ jobs:
     - name: Run container build script
       run: bin/build_container_image
       env:
+        TRAVIS_BRANCH: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


Fryguy:
I think it should be ref_name... That's what I was checking in my pr

.github/workflows/ci.yaml
```diff
@@ -32,5 +32,5 @@ jobs:
     - name: Run container build script
       run: bin/build_container_image
       env:
+        TRAVIS_BRANCH: ${{ github.ref }}
```
Fryguy, line 35: ref won't work because ref == "refs/heads/master" on a regular branch.

Fryguy:
I opened #240 with my branch (we were doing it at the same time)

bdunne:
Closing in favor of #240

PR #240, opened 2021-12-20
Fix TRAVIS env vars

Fryguy:
@bdunne Please review.  Alternative to #239 using the techniques in https://developer.okta.com/blog/2020/05/18/travis-ci-to-github-actions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Set TRAVIS ENV variables
+      run: |
+        echo "TRAVIS_PULL_REQUEST=${{ github.event.number || false }}" >> $GITHUB_ENV
+        TRAVIS_BRANCH="${{ github.head_ref }}" && \
+          echo "TRAVIS_BRANCH=${TRAVIS_BRANCH:-$(echo $GITHUB_REF | cut -d '/' -f 3-)}" >> $GITHUB_ENV
     - name: Set up registry credentials
       if: ${{ github.event_name != 'pull_request' }}
       run: |
@@ -31,6 +36,3 @@ jobs:
         echo "REGISTRY_PASSWORD=${{ secrets.REGISTRY_PASSWORD }}" >> $GITHUB_ENV
     - name: Run container build script
       run: bin/build_container_image
-      env:
-        TRAVIS_PULL_REQUEST: ${{ github.event_name == 'pull_request' && 'true' || 'false' }}
-        TRAVIS_BRANCH: ${{ env.GITHUB_REF_NAME }}


.github/workflows/ci.yaml
```diff
@@ -22,15 +22,18 @@ jobs:
   build:
     needs: ci
     runs-on: ubuntu-latest
+    env:
+      TRAVIS_BRANCH: ${{ github.head_ref }}
+      TRAVIS_PULL_REQUEST: ${{ github.event.number || 'false' }}
```
Fryguy, line 27: This one I'm not 100% sure about, because I don't know if github.event.number returns a falsey value like this.

.github/workflows/ci.yaml
```diff
@@ -22,15 +22,18 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
+    - name: Set ENV variables
+      run: |
+        echo "::set-env name=TRAVIS_BRANCH::${TRAVIS_BRANCH:-$(echo $GITHUB_REF | cut -d '/' -f 3)}"
```
Fryguy, line 32: This is annoying, but apparently it's the only way to get the actual branch name.

.github/workflows/ci.yaml
Fryguy: The article says it should return false by default, so I'm leaving it like in the article.

.github/workflows/ci.yaml
```diff
@@ -22,15 +22,18 @@ jobs:
   build:
     needs: ci
     runs-on: ubuntu-latest
+    env:
+      TRAVIS_BRANCH: ${{ github.head_ref }}
```
Fryguy, line 26: On a PR this will be the branch name, on a non-PR, this will be blank, but then set later in the Set ENV variables section.

.github/workflows/ci.yaml
```diff
@@ -22,15 +22,18 @@ jobs:
...
+      TRAVIS_PULL_REQUEST: ${{ github.event.number }}
```
Fryguy, line 27: I moved this up to the global section only because we have the other one here. I'd rather leave it in the specific step, but then it's defined in 2 places, which is gross.

.github/workflows/ci.yaml
bdunne: What do you think of removing this and just using the `Set ENV variables` step?  Then it be `master` or `240`, right?

Fryguy: Let me see if I can play with that - I'm playing around in https://github.com/Fryguy/sandbox

Fryguy: Done

Fryguy:
Backported to `morphy` in commit ae49f8e35f88bc89987fa4fb780681dae22ab9ed.

```text
commit ae49f8e35f88bc89987fa4fb780681dae22ab9ed
Date:   Mon Dec 20 15:10:44 2021 -0500

    Merge pull request #240 from Fryguy/fix_travis_env
    
    Fix TRAVIS env vars
    
    (cherry picked from commit 44fbe2076e710d5106fb37a99b27589fa0843996)
```

PR #241, opened 2021-12-20
Remove TRAVIS env vars entirely

Fryguy:


bin/build_container_image
```diff
@@ -3,7 +3,11 @@
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
 DEFAULT_TAG="latest"
-[[ $TRAVIS_BRANCH != "master" ]] && DEFAULT_TAG="$DEFAULT_TAG-$TRAVIS_BRANCH"
+if [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
+  DEFAULT_TAG="$DEFAULT_TAG-pr"
```
Fryguy, line 7: @bdunne Thoughts here?  SInce the actual name doesn't matter, do we even care about this?  I was thinking we could just leave it as "latest"

bin/build_container_image
```diff
@@ -12,7 +16,7 @@ set -e
 
 docker build . -t $IMAGE_NAME
 
-[[ $TRAVIS_PULL_REQUEST != "false" ]] && exit 0
+[[ -z "$REGISTRY_USERNAME" ]] && exit 0
```
Fryguy, line 19: I thought it simpler to just check for the presence of the REGISTRY_USERNAME, as opposed to checking for anything CI specific

bin/build_container_image
Fryguy: I noticed, previously that if run locally this would generate `latest-` as the name, because TRAVIS_BRANCH was blank  (actually it would probably blow up if run locally)...thinking I might fix that local use case as well

Fryguy: I updated with this and fixed the `latest-` thing.  Also fixed it so that branches with `/` will work (even though that doesn't effect master, it could affect pushes to forks, where the fork has the push event.

bin/build_container_image
```diff
@@ -3,7 +3,8 @@
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
 DEFAULT_TAG="latest"
-[[ $TRAVIS_BRANCH != "master" ]] && DEFAULT_TAG="$DEFAULT_TAG-$TRAVIS_BRANCH"
+[ "$GITHUB_EVENT_NAME" == "push" -a "$GITHUB_REF" != "refs/heads/master" ] && \
+  DEFAULT_TAG="$DEFAULT_TAG-${GITHUB_REF_NAME/\//-}"
```
Fryguy, line 7:
```bash
$ GITHUB_EVENT_NAME=push GITHUB_REF_NAME="master" bin/build_container_image
docker.io/manageiq/rpm_build:latest-master
$ GITHUB_EVENT_NAME=push GITHUB_REF_NAME="foobar" bin/build_container_image
docker.io/manageiq/rpm_build:latest-foobar
$ GITHUB_EVENT_NAME=push GITHUB_REF_NAME="feature/foobar" bin/build_container_image
docker.io/manageiq/rpm_build:latest-feature-foobar
$ bin/build_container_image
docker.io/manageiq/rpm_build:latest
```

Fryguy:
@bdunne This is ready to go.

bin/build_container_image
Fryguy: This PR is an example where it will tag as latest: https://github.com/ManageIQ/manageiq-rpm_build/runs/4587369238?check_suite_focus=true#step:4:3079

bin/build_container_image
bdunne: What do you think of switching to `latest-master`so that we can get rid of this conditional logic?

bin/build_container_image
bdunne: I'm hesitant to add `GITHUB_EVENT_NAME`, maybe we can get the ref from git itself?

bin/build_container_image
Fryguy: It's super common to have latest be the latest, so I'm hesitant to _not_ use latest now that we've already started.

bin/build_container_image
Fryguy: I can try, but we need the even to know if it's a PR or a push build...not sure how you get that out of git.  Basically, we need push builds that are not master (i.e. morphy, lasker, etc.)

PR #243, opened 2022-01-15
Changes made to support s390x platform

dpkulkarni:


Dockerfile
```diff
@@ -4,8 +4,7 @@ ENV TERM=xterm \
     APPLIANCE=true \
     RAILS_USE_MEMORY_STORE=true
 
-RUN dnf -y remove *subscription-manager* && \
```
bdunne, line 7: Why is this being removed?  Are you planning to use subscription-manager?

Dockerfile
```diff
@@ -16,8 +15,7 @@ RUN dnf -y remove *subscription-manager* && \
       https://rpm.manageiq.org/release/14-najdorf/el8/noarch/manageiq-release-14.0-1.el8.noarch.rpm && \
     dnf -y module enable ruby:2.7 && \
     dnf -y module enable nodejs:12 && \
-    if [ ${ARCH} != "s390x" ] ; then dnf -y module disable virt:rhel; fi && \
-    dnf config-manager --setopt=ubi-8-*.exclude=rpm* --save && \
```
bdunne, line 20: Do we want to pull rpm* packages from the UBI repos?  This has caused build failures in the past where there is a newer version available in the UBI repos that has a dependency that is not yet available in the upstream CentOS repos.  Have you tested this on PPC and x86_64 to verify that the upstream and downstream are now in sync?

dpkulkarni:
Adding @sriramasharma in loop

Dockerfile
Fryguy: This seems to be needed to pull from a private internal yum repository.  I think we can add this behind a conditional for now.

Dockerfile
Fryguy: I left this line in and it worked fine, so we can undo this line.

bdunne: We may need a conditional here.  When building on RHEL (and for s390x where there is no CentOS repo) we probably don't want this exclude.

PR #246, opened 2022-01-24
Ensure that the bundler version in the script matches the container

bdunne:
This will put the correct version in the gemset GEM_HOME


Fryguy:
I think this particular line puts bundler into the gemset as opposed to the one from the system, so it gets delivered in the RPM.  And then the build uses this particular bundler. I think this line has to stay.

Fryguy:
Backported to `morphy` in commit 82af4a74d2926750ab71276ceccab7036985dd0c.

```text
commit 82af4a74d2926750ab71276ceccab7036985dd0c
Date:   Mon Jan 24 11:21:14 2022 -0500

    Merge pull request #246 from bdunne/bundler_reinstall
    
    Ensure that the bundler version in the script matches the container
    
    (cherry picked from commit 769553a3276de66217bb55b2d1274428e3f6e553)
```

PR #268, opened 2022-02-18
Build the ARCH and Source RPMs

bdunne:
SRPMs are hepful when we need to create hotfixes since they contain sources and the RPM spec, then we can add any patches and rebuild the RPMs.
Otherwise we need to start over with the whole tar generation process (including bundling the gems)

This results in:
```
Wrote: /root/BUILD/rpms/x86_64/manageiq-13.1.0-20220218221933.el8.src.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-appliance-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-appliance-tools-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-core-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-gemset-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-pods-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-system-13.1.0-20220218221933.el8.x86_64.rpm
Wrote: /root/BUILD/rpms/x86_64/manageiq-ui-13.1.0-20220218221933.el8.x86_64.rpm
```

### DIFF
--- a/lib/manageiq/rpm_build/build_copr.rb
+++ b/lib/manageiq/rpm_build/build_copr.rb
@@ -26,7 +26,8 @@ module ManageIQ
             shell_cmd("rpmbuild -bs --define '_sourcedir #{RPM_SPEC_DIR}' --define '_srcrpmdir #{RPM_SPEC_DIR}' #{rpm_spec}")
             shell_cmd("copr-cli build -r epel-8-x86_64 #{rpm_repo_name} #{OPTIONS.product_name}-*.src.rpm")
           else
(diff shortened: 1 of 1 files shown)


bdunne:
@Fryguy Any thoughts on this?

Fryguy:
Oh yes, I love this.  Only question I have is how this will affect builds, but I think we can handle that in a follow up.

PR #283, opened 2022-06-21
Add more manifests for non-gem repos

Fryguy:
This PR creates a copy of the BUILD file that's normally in /var/www/miq/vmdb in the /opt/manageiq/manifests directory, and then adds additional BUILD_ files for non-gem repos that we normally don't get any information for.  gem repos and their SHAs already exist in /opt/manageiq/manifests/gen_manifest.csv.

@jrafanie Please review.

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -9,11 +9,13 @@ module ManageIQ
     class GenerateCore
       include Helper
 
-      attr_reader :miq_dir, :ui_service_dir
+      attr_reader :miq_dir, :ui_service_dir, :appliance_dir, :manifest_dir
 
       def initialize
-        @miq_dir = BUILD_DIR.join("manageiq")
+        @miq_dir        = BUILD_DIR.join("manageiq")
         @ui_service_dir = BUILD_DIR.join("manageiq-ui-service")
+        @appliance_dir  = BUILD_DIR.join("manageiq-appliance")
+        @manifest_dir   = MANIFEST_DIR
       end
 
       def build_file
@@ -85,6 +87,7 @@ module ManageIQ
         build_service_ui
         seed_ansible_runner
         compile_locale_files
+        generate_manifests
 
         if OPTIONS.npm_registry
           Dir.chdir(miq_dir) do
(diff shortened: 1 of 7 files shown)


lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -161,6 +164,20 @@ def compile_locale_files
...
       end
+
+      def generate_manifests
+        FileUtils.mkdir_p(manifest_dir)
+
+        [
+          miq_dir,        "BUILD",
+          appliance_dir,  "BUILD_APPLIANCE",
+          ui_service_dir, "BUILD_UI_SERVICE",
+          ROOT_DIR,       "BUILD_RPM_BUILD"
```
Fryguy, line 175: I originally built these all into a single csv file, but then realized that I wanted to split them by RPM.  Maybe it's overkill, but this is why I chose this path.

lib/manageiq/rpm_build/generate_gemset.rb
```diff
@@ -109,8 +109,7 @@ def scrub
       def generate_dependency_manifest
         where_am_i
 
-        FileUtils.rm_rf(manifest_dir) if manifest_dir.exist?
-        FileUtils.mkdir(manifest_dir)
+        FileUtils.mkdir_p(manifest_dir)
```
Fryguy, line 112: I'm not really sure of the purpose of removing the original manifest dir.

lib/manageiq/rpm_build/generate_core.rb
Fryguy: I can change from ALL_CAPS to something else.  This started as an extension to the existing BUILD file, but these will all live in the manifest dir instead.

Fryguy: @jrafanie or maybe @agrare ...would like your opinions on this.  Separate files split into RPMs, or just a single CSV in core rpm?  Or something else?

jrafanie: Is the manifest for each the same as `rpm -ql ...` to list the files in the rpm?  If so, yes, I agree, manifests should match the rpm structure so 1 per rpm.

Fryguy: These manifest files are basically the git shas of the repos for diagnostic... Nothing to do with rpm -q.

jrafanie: Can you post an example?  I'm not sure what options are being discussed.

Fryguy:
```
$ cat BUILD
20220622112412_abcd1234
$ cat BUILD_APPLIANCE
20220622112412_1234abcd
...
```

vs

```csv
name,git_sha
manageiq,abcd1234
manageiq-appliance,1234abcd
...
```

jrafanie: discussed on a call... it makes sense to keep the references in the rpm subpackages clean and simple so I like the idea of separate files... your first option here.

jrafanie (approved):
LGTM

Fryguy:
Backported to `najdorf` in commit 2aa3bd0961980f6667916f9dcb4cafc8fcba3eae.

```text
commit 2aa3bd0961980f6667916f9dcb4cafc8fcba3eae
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Thu Jun 23 17:38:02 2022 -0400

    Merge pull request #283 from Fryguy/more_manifest
    
    (cherry picked from commit d06a5282563251d07d8f2821134d7cd4f7263012)
```

PR #294, opened 2022-08-22
[oparin] Bundle without development and test in production

bdunne:
Also, sort alphabetically


lib/manageiq/rpm_build/generate_gemset.rb
```diff
@@ -64,8 +64,9 @@ def populate_gem_home(build_type)
             shell_cmd("bundle config set build.unf_ext --with-cxxflags='-fsigned-char'")
           end
 
-          shell_cmd("bundle config set --local with qpid_proton systemd")
           shell_cmd("bundle config set --local build.rugged --with-ssh")
+          shell_cmd("bundle config set --local with qpid_proton systemd")
```
Fryguy, line 68: Since you made appliance optional, I think you need to include it here too.

Fryguy:
Let's make this a direct oparin PR similar to https://github.com/ManageIQ/manageiq/pull/22067, then we can forward port it all together.

PR #295, opened 2022-08-22
Not really sure why this is here, we specify ~>3.0

bdunne:
https://github.com/ManageIQ/manageiq/blob/a475afa6e8e834e7332b612924fc6f47405f8454/Gemfile#L84

### DIFF
--- a/lib/manageiq/rpm_build/generate_gemset.rb
+++ b/lib/manageiq/rpm_build/generate_gemset.rb
@@ -57,7 +57,6 @@ module ManageIQ
 
         Dir.chdir(miq_dir) do
           shell_cmd("gem env")
-          shell_cmd("gem install mime-types -v 2.6.1")
 
           if RUBY_PLATFORM.match?(/powerpc64le|s390x/)
             shell_cmd("bundle config set build.sassc --disable-march-tune-native")


kbrock:
When you are building rpms, there are 2 different `Gemfile`s in play. right?
Not sure if this is using the `Gemfile` that you are referencing.

Having said that, I do think that dropping the version is a good idea.

Have you had a chance to build rpms with this change?

bdunne:
> When you are building rpms, there are 2 different `Gemfile`s in play. right? Not sure if this is using the `Gemfile` that you are referencing.
> 
> Having said that, I do think that dropping the version is a good idea.
> 
> Have you had a chance to build rpms with this change?

I haven't built rpms with this change yet, it's on my TODO list.

Fryguy:
> When you are building rpms, there are 2 different Gemfiles in play. right?

Which 2 are you referring to?  I thought there is on the one in core?

bdunne:
Images built and installed on an appliance successfully.  I can log in ot the UI, is there anything else that I need to test?

kbrock:
> > When you are building rpms, there are 2 different Gemfiles in play. right?
> 
> Which 2 are you referring to? I thought there is on the one in core?

There is a Gemfile for the local scripts, and a Gemfile inside the docker conatainer from miq core. For a while, they had different dependencies and were even running a different version of ruby/bundler

PR #293, opened 2022-08-18
Server certs need to be group readable

agrare:
The server certificate is used by httpd (running as root) and MiqRemoteConsoleWorker (running as manageiq).
This fixes existing certs on upgrade, for certs created on first boot see https://github.com/ManageIQ/manageiq-appliance/pull/365/

https://github.com/ManageIQ/manageiq/issues/22065

PR #298, opened 2022-08-25
[WIP] make old certs readable by manageiq user

kbrock:
### Background

The ssl certificates are created outside the rpm
Older installations had root own all files.
So there were no permissions issues.

Now (for some time), most processes are run by manageiq user.
So to make things work better for the end user, we change the
permissions of critical (and sensitive) files to be restrictive
but readable by manageiq.manageiq

### Bug

v2_key was converted to be read by `manageiq` user, but ssl certificates were not.
Webconsole is broken because it needs to read the ssl certificate key.

### Solution

Make all certificates readable by `manageiq` user.
Issue raised in https://github.com/ManageIQ/manageiq/discussions/22060

### Limitations

Assumes that there will not be an ssl key without a `v2_key`.
Since nothing works without a `v2_key`, we can choose this single file to detect.
reminder: `[ -e certs/* ]` generates an error if there is more than one file in the directory: `-bash: [: too many arguments` error.

### DIFF
--- a/rpm_spec/subpackages/manageiq-core
+++ b/rpm_spec/subpackages/manageiq-core
@@ -35,8 +35,8 @@ done
 # These directories contain files not owned by this rpm.
 #  For upgrades, ensure the files have the correct group privs
 #  so root and manageiq users can read them.
-[ -e %{app_root}/certs/v2_key ] && %{__chown} manageiq.manageiq %{app_root}/certs/v2_key
-[ -e %{app_root}/certs/v2_key ] && %{__chmod} o-rw %{app_root}/certs/v2_key
+[ -e %{app_root}/certs/v2_key ] && %{__chown} manageiq.manageiq %{app_root}/certs/*
+[ -e %{app_root}/certs/v2_key ] && %{__chmod} o-rw %{app_root}/certs/*
 
 %{__chown} -R manageiq.manageiq %{app_root}/log
 %{__chmod} -R o-rw %{app_root}/log


kbrock:
WIP: need to build rpms and see that this does not blow up / does indeed chmod the ssl certs and v2 key from root to `manageiq.manageiq`

rpm_spec/subpackages/manageiq-core
```diff
@@ -35,8 +35,8 @@ done
 # These directories contain files not owned by this rpm.
 #  For upgrades, ensure the files have the correct group privs
 #  so root and manageiq users can read them.
-[ -e %{app_root}/certs/v2_key ] && %{__chown} manageiq.manageiq %{app_root}/certs/v2_key
-[ -e %{app_root}/certs/v2_key ] && %{__chmod} o-rw %{app_root}/certs/v2_key
+[ -e %{app_root}/certs/v2_key ] && %{__chown} manageiq.manageiq %{app_root}/certs/*
+[ -e %{app_root}/certs/v2_key ] && %{__chmod} o-rw %{app_root}/certs/*
```
agrare, line 39: Will this cause apache to not be able to read the server.cer?

kbrock: I'm pretty sure Apache is more concerned with the selinux privileges.
It should be running as `root` at this time so should be able to read it.

But yea, that is why this is WIP.

kbrock: Also, when the selinux privileges are off, then it is confusing because it is running as `root` and the error messages look like a `chmod` issue. `ls -Z certs` and `restorecon -R certs` is our friend when this happens.

kbrock:
looks like this has been fixed by https://github.com/ManageIQ/manageiq-rpm_build/pull/293

PR #299, opened 2022-08-26
LicenseFinder ignore dev and test bundler groups

bdunne:
Now that conditionals have been removed from the core Gemfile, all dependencies are listed in the Gemfile.lock.  License Finder does not load the bundler "without", so we have to tell it to also ignore those dependencies otherwise it will try to materialize those gems (which are not installed)

Example error:
```
gems/bundler-2.3.21/lib/bundler/definition.rb:520:in `materialize': Could not find foreman-0.87.2, manageiq-style-1.2.0, PoParser-3.2.6, ... in locally installed gems (Bundler::GemNotFound)
```

### DIFF
--- a/lib/manageiq/rpm_build/generate_gemset.rb
+++ b/lib/manageiq/rpm_build/generate_gemset.rb
@@ -138,6 +138,9 @@ module ManageIQ
       def run_license_finder(repos, type)
         where_am_i
 
+        shell_cmd("license_finder ignored_groups add development")
+        shell_cmd("license_finder ignored_groups add test")
+
         output  = manifest_dir.join("#{type}_manifest.csv")
         columns = "name version licenses"
         shell_cmd("license_finder report --format csv --write-headers --aggregate-paths #{repos} --columns #{columns} --save #{output}")


bdunne:
RPM builds are broken due to the License Finder errors.
Caused by removing the Gemfile conditionals in https://github.com/ManageIQ/manageiq/pull/22073

lib/manageiq/rpm_build/generate_gemset.rb
```diff
@@ -138,6 +138,9 @@ def generate_npm_manifest
       def run_license_finder(repos, type)
         where_am_i
 
+        shell_cmd("license_finder ignored_groups add development")
+        shell_cmd("license_finder ignored_groups add test")
```
jrafanie, line 142: Dumb question.  Does this work? I don't see other calls to `license_finder` with `ignored_groups`.  Gemfile.lock manipulation is tricky so want to make sure that works as we want.

jrafanie: I guess the change that we're trying to workaround will break systems with the rpms installed if you run bundle as it will fully resolve the Gemfile and look for the "ignored" dependencies in order to resolve the dependency tree even if it doesn't need to install/use them. 

I don't know if that is a new problem.

bdunne: I ran a build locally and it works.  Build succeeds, manifest is generated, dev dependencies are not in the resulting RPMs.

bdunne: I don't think running `bundle install` on a RPM system was ever supported.  I think it will work because we ship the Gemfile.lock and the bundler config in the RPMs, so it should be a no-op.

jrafanie (approved):
LGTM.

PR #300, opened 2022-08-31
Don't add anything in rpm_cache directory to the docker image

bdunne:
This could cause the image to be enormous


Fryguy:
What is the rpm_cache used for?

bdunne:
> What is the rpm_cache used for?

In order to regenerate the RPM repo metadata, we need the RPMs downloaded locally.  This is the directory where they are stored.

PR #303, opened 2022-09-19
Add a nightly rpm build script

bdunne:
TODO:
- [x] Credentials for registry and RPM Repo
- [x] Trigger pods build
- ~~[ ] Tag builds~~ I think we'll handle tag builds manually on hte build machine as we have in the past?

### DIFF
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -29,3 +29,17 @@ jobs:
       run: |
         echo $REGISTRY_PASSWORD | docker login docker.io --password-stdin --username $REGISTRY_USERNAME
         docker run docker.io/manageiq/rpm_build:latest build --build-type nightly
+  notify_builders:
+    needs: build_rpms
+    runs-on: ubuntu-latest
+    steps:
+    - name: Notify manageiq-pods on RPM build
+      uses: peter-evans/repository-dispatch@v1
+      with:
+        token: ${{ secrets.DEPLOY_TOKEN }}
+        repository: ManageIQ/manageiq-pods
+        event-type: build
+        client-payload: |
+          { "repository": "${{ github.repository }}",
+            "ref_name": "${{ github.ref_name }}",
+            "sha": "${{ github.sha }}" }
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -7,11 +7,11 @@ on:
 
 jobs:
   build_rpm_build_container:
+    if: github.repository_owner == 'ManageIQ'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
     - name: Set up registry credentials
-      if: ${{ github.event_name != 'pull_request' }}
       run: |
         echo "REGISTRY_USERNAME=${{ secrets.REGISTRY_USERNAME }}" >> $GITHUB_ENV
         echo "REGISTRY_PASSWORD=${{ secrets.REGISTRY_PASSWORD }}" >> $GITHUB_ENV
@@ -19,6 +19,7 @@ jobs:
       run: bin/build_container_image
   build_rpms:
     needs: build_rpm_build_container
+    if: github.repository_owner == 'ManageIQ'
     runs-on: ubuntu-latest
     steps:
     - name: Set up registry credentials
@@ -38,6 +39,7 @@ jobs:
         docker run docker.io/manageiq/rpm_build:latest build -v options:/root/OPTIONS --build-type nightly
   notify_builders:
     needs: build_rpms
+    if: github.repository_owner == 'ManageIQ'
     runs-on: ubuntu-latest
     steps:
     - name: Notify manageiq-pods on RPM build
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -1,0 +1,31 @@
+name: Build RPMs
+
+on:
+  schedule:
+  - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build_rpm_build_container:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up registry credentials
+      if: ${{ github.event_name != 'pull_request' }}
+      run: |
+        echo "REGISTRY_USERNAME=${{ secrets.REGISTRY_USERNAME }}" >> $GITHUB_ENV
+        echo "REGISTRY_PASSWORD=${{ secrets.REGISTRY_PASSWORD }}" >> $GITHUB_ENV
+    - name: Run container build script
+      run: bin/build_container_image
+  build_rpms:
+    needs: build_rpm_build_container
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up registry credentials
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,55 @@
...
+  schedule:
+  - cron: '0 0 * * *'
+
+jobs:
+  build_rpm_build_container:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        repository: ManageIQ/manageiq-rpm_build
```
Fryguy, line 14: Don't think you need this as it's implied.

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,55 @@
+name: Build RPMs
+
+on:
+  # push:
+  schedule:
```
Fryguy, line 5: we probably also want workflow_dispatch so we can build manually on demand

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        repository: ManageIQ/manageiq-rpm_build
+    - name: Set up registry credentials
+      if: ${{ github.event_name != 'pull_request' }}
+      run: |
+        echo "REGISTRY_USERNAME=${{ secrets.REGISTRY_USERNAME }}" >> $GITHUB_ENV
+        echo "REGISTRY_PASSWORD=${{ secrets.REGISTRY_PASSWORD }}" >> $GITHUB_ENV
```
Fryguy, line 19: Do we need a remote registry?  I thought docker build support was built in and I think you have a local registry, which is all we need here.

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Build and push RPMs
+      run: |
+        echo $REGISTRY_PASSWORD | docker login docker.io --password-stdin --username $REGISTRY_USERNAME
+        mkdir options
+        echo "---" > options/options.yml
+        echo "rpm_repository:" >> options/options.yml
+        echo "  digitalocean_access_token: ${{ secrets.RPM_BUILD_DO_ACCESS_TOKEN }}" >> options/options.yml
+        echo "  s3_api:" >> options/options.yml
+        echo "    access_key: ${{ secrets.RPM_BUILD_S3_ACCESS_KEY " >> options/options.yml
+        echo "    secret_key: ${{ secrets.RPM_BUILD_S3_SECRET_KEY " >> options/options.yml
```
Fryguy, line 39: I think there's a slightly cleaner way to do this, but this is fine for now.  I do worry that it will expose the secrets like if the options.yml is printed or something.

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+    - name: Notify manageiq-pods on RPM build
+      uses: peter-evans/repository-dispatch@v1
+      with:
+        token: ${{ secrets.DEPLOY_TOKEN }}
+        repository: ManageIQ/manageiq-pods
+        event-type: build
+        client-payload: |
+          { "repository": "${{ github.repository }}",
+            "pr_number": "${{ steps.regex-commit-message.outputs.group1 }}",
```
Fryguy, line 53: This doesn't exist here - I assume you copied it from the website build, but you're missing the steps where it looks at the HEAD commit message to pull out the PR number: https://github.com/ManageIQ/manageiq.org/blob/master/.github/workflows/ci.yaml#L26-L32

.github/workflows/build_rpms.yaml
bdunne: There's just the local copy on the machine it was built on.  If jobs run on different machines they won't have access to the image that was built.  Also, I see little downside to pushing the image to docker.io especially for troubleshooting purposes.

.github/workflows/build_rpms.yaml
bdunne: Yeah, we either have to write a file somewhere or modify the code to accept ENV variables or something.

.github/workflows/build_rpms.yaml
bdunne: Yes, it was copied.  I didn't know what we want to put in here yet.

.github/workflows/build_rpms.yaml
bdunne: Done

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build_rpm_build_container:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up registry credentials
+      if: ${{ github.event_name != 'pull_request' }}
```
Fryguy, line 14: I think this isn't quite right because this will also run on branches pushed to local forks.  We probably want something that limits this only to the official repo? or perhaps write it in such a way that the code runs anywhere, but only executes if the values are set... `if [ -n "${{ secrets.REGISTRY_USERNAME }}"]; then`...

.github/workflows/build_rpms.yaml
Fryguy: I'd prefer env vars moving forward - never really liked secrets in the options.

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,52 @@
...
+  notify_builders:
+    needs: build_rpms
+    runs-on: ubuntu-latest
+    steps:
+    - name: Notify manageiq-pods on RPM build
+      uses: peter-evans/repository-dispatch@v1
+      with:
+        token: ${{ secrets.DEPLOY_TOKEN }}
+        repository: ManageIQ/manageiq-pods
+        event-type: build
```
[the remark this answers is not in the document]
bdunne: https://github.com/ManageIQ/manageiq-pods/pull/889

.github/workflows/build_rpms.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Build and push RPMs
+      run: |
+        echo $REGISTRY_PASSWORD | docker login docker.io --password-stdin --username $REGISTRY_USERNAME
+        mkdir options
+        echo "---" > options/options.yml
+        echo "rpm_repository:" >> options/options.yml
+        echo "  digitalocean_access_token: ${{ secrets.RPM_BUILD_DO_ACCESS_TOKEN }}" >> options/options.yml
+        echo "  s3_api:" >> options/options.yml
+        echo "    access_key: ${{ secrets.RPM_BUILD_S3_ACCESS_KEY " >> options/options.yml
+        echo "    secret_key: ${{ secrets.RPM_BUILD_S3_SECRET_KEY " >> options/options.yml
```
Fryguy, line 38: @bdunne Here's the bug causing the failure in https://github.com/ManageIQ/manageiq-rpm_build/actions/runs/3277198132

```suggestion
        echo "    access_key: ${{ secrets.RPM_BUILD_S3_ACCESS_KEY }}" >> options/options.yml
        echo "    secret_key: ${{ secrets.RPM_BUILD_S3_SECRET_KEY }}" >> options/options.yml
```

PR #311, opened 2022-10-03
Sync python updates

Fryguy:
@agrare Please review.

This packages hasn't ever shipped, but it does create whitesource issues.  This PRs syncs the list with what we have in manageiq-appliance-build.


agrare (approved):

agrare:
@Fryguy are the [new vulnerabilities](https://github.com/ManageIQ/manageiq-rpm_build/pull/311/checks?check_run_id=8685621304) something you want to address here or later?

Fryguy:
oauthlib should have been handled by this....let me fix that.
ipaddress is known and will have to be a follow up

Fryguy:
@agrare Let's merge as is...this moves us forward, oauthlib should have already been an open issue, and in practice, oauthlib 3.2.1 is the one that gets installed anyway:

```
$ grep oauthlib /var/lib/manageiq/installed.txt
oauthlib==3.2.1
requests-oauthlib==1.3.1
```

PR #314, opened 2022-10-14
Update building of manageiq-ansible-venv

Fryguy:
@bdunne Please review.  I verified locally this creates the RPM and it installs the correct content (though there are things we can cleanup later in a followup)

I will have follow ups later that require this RPM from the other RPMs, but I'm also wondering if maybe this should be a subpackage of the other RPMs instead.  Also something we can decide later.

### DIFF
--- a/packages/manageiq-ansible-venv/Dockerfile
+++ b/packages/manageiq-ansible-venv/Dockerfile
@@ -1,7 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi
 
-ARG ARCH=x86_64
-
 # For productization
 ENV ORG_NAME=manageiq \
     PRODUCT_NAME=manageiq \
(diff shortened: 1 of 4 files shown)


packages/manageiq-ansible-venv/container-assets/build.sh
```diff
@@ -6,36 +6,42 @@ venv_path=$VENV_ROOT/venv
...
 sed -i "s/VERSION/$VERSION/" $RPM_SPEC
 sed -i "s#VENV_ROOT#$VENV_ROOT#" $RPM_SPEC
 
-if [ -f "/root/.config/copr" ]; then
-  rpmbuild -bs --define "_sourcedir $rpm_path" --define "_srcrpmdir $rpm_path" /$RPM_SPEC
-  copr-cli build -r $COPR_CHROOT $COPR_PROJECT $rpm_path/*.src.rpm
-else
-  rpmbuild -bb --define "_sourcedir $rpm_path" --define "_rpmdir $rpm_path" /$RPM_SPEC
-fi
+rpmbuild -bb --define "_sourcedir $rpm_path" --define "_rpmdir $rpm_path" /$RPM_SPEC
```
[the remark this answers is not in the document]
Fryguy: I think some of the packages are compiled, so yes, I think this one will be arch specific.

packages/manageiq-ansible-venv/Dockerfile
```diff
@@ -14,27 +14,25 @@ ENV RPM_SPEC=manageiq-ansible-venv.spec \
 # CentOS repos: rpm-build and copr-cli dependencies
 # EPEL: copr-cli
 RUN sed -i 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/subscription-manager.conf
-RUN dnf -y install \
+RUN dnf config-manager --setopt=tsflags=nodocs --save && \
```
bdunne, line 17: We should probably set ARCH here like we do in the other images rather than expecting a build arg.
https://github.com/ManageIQ/manageiq-rpm_build/blob/8cc957a8dd8d2068335a64c65361156e3e2f72f8/Dockerfile#L7

packages/manageiq-ansible-venv/Dockerfile
```diff
@@ -14,27 +14,25 @@ ENV RPM_SPEC=manageiq-ansible-venv.spec \
...
+    dnf -y install \
       http://mirror.centos.org/centos/8-stream/BaseOS/${ARCH}/os/Packages/centos-stream-repos-8-2.el8.noarch.rpm \
       http://mirror.centos.org/centos/8-stream/BaseOS/${ARCH}/os/Packages/centos-gpg-keys-8-2.el8.noarch.rpm \
       https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
-    dnf config-manager --setopt=tsflags=nodocs --save && \
     dnf -y install \
-      copr-cli \
+      make \
       gcc \
+      krb5-devel \
```
bdunne, line 25: Interesting, what required this?

Fryguy: I matched this list to the one in the python_modules.ks.erb.  My plan was to try to take these out 1 by 1, but I wanted them to match 1 for 1 with what was there.

bdunne (commented):
I think the ARCH is the only thing we should do right away.

packages/manageiq-ansible-venv/Dockerfile
```diff
@@ -14,27 +14,25 @@ ENV RPM_SPEC=manageiq-ansible-venv.spec \
 # CentOS repos: rpm-build and copr-cli dependencies
 # EPEL: copr-cli
 RUN sed -i 's/enabled=1/enabled=0/' /etc/yum/pluginconf.d/subscription-manager.conf
```
Fryguy, line 16: @bdunne I noticed this line - should I make it match how we do it in the other images?

https://github.com/ManageIQ/manageiq-rpm_build/blob/8cc957a8dd8d2068335a64c65361156e3e2f72f8/Dockerfile#L8

bdunne: probably, the consistency would be nice

Fryguy:
Ok updated @bdunne - I also removed EPEL because that was only added for copr-cli

PR #326, opened 2022-10-25
[WIP] Build qpid-proton v0.37 for EL9

bdunne:
This is a copy of https://src.fedoraproject.org/rpms/qpid-proton epel9 branch.


agrare:
@bdunne were we using the qpid-proton rpm from the el8 repos previously?

agrare:
> Do we have a way to validate that this upgrade doesn't break anything?

We will have to track down a full-up openstack (not microstack) with STF event service

bdunne:
> @bdunne were we using the qpid-proton rpm from the el8 repos previously? Just asking because this looks like a net-new rpm not just an upgrade

I think the previous build had been done in COPR, so the spec and patches were not in this repo.  In the first commit, I opened the SRPM and committed those files (v0.30.0).  The second commit is the upgrade to v0.37.0.

agrare:
@bdunne let's start with bumping to 0.37.0 in openstack https://github.com/ManageIQ/manageiq-providers-openstack/pull/830 and nuage and let the specs run

agrare:
@bdunne do we still need this for EL9?

jrafanie:
@kbrock let me know about this PR... we also need qpid-proton 0.37.0+ built for ruby 3.1 support...

from https://github.com/ManageIQ/manageiq/pull/22698

```
    Upgrade qpid_proton to 0.37.0 for ruby 3.1 support

    Fixes an issue in their extension of ruby's URI library which
    was broken when they used private class instance variable schemes which
    is changed in ruby 3.1.

    Fixed in 0.37.0 of qpid-proton:
    https://github.com/apache/qpid-proton/commit/6bcdfebb55ea3554bc29b1901422532db331a591
    https://github.com/apache/qpid-proton/commit/ec535e772e02bd67a34a90d301c450758f53833e
```

jrafanie:
@bdunne did we drop epel?  I see qpid proton 0.37.0 in epel9 https://mirror.steadfastnet.com/epel/9/Everything/x86_64/Packages/q/

bdunne:
Looks like this was backported to EPEL8, so we don't need to ship this RPM ourselves.

PR #331, opened 2022-11-01
Fix missing packages in manageiq-ansible-venv

Fryguy:
Followup fixes after https://github.com/ManageIQ/manageiq-appliance-build/issues/423

This fixes issues where psutil was missing from the system site-packages.  After this PR, I can successfully run playbooks (tested with the hello world and vsphere playbooks)

@bdunne or @jrafanie Please review.

### DIFF
--- a/lib/manageiq/rpm_build/generate_ansible_venv.rb
+++ b/lib/manageiq/rpm_build/generate_ansible_venv.rb
@@ -46,7 +46,10 @@ module ManageIQ
       def install_python_system_packages
         where_am_i
 
-        shell_cmd("#{pip_versioned} install virtualenv psutil==5.6.6", "PYTHONDONTWRITEBYTECODE" => "1")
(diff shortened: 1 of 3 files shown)


lib/manageiq/rpm_build/generate_ansible_venv.rb
```diff
@@ -46,7 +46,10 @@ def populate
       def install_python_system_packages
         where_am_i
 
-        shell_cmd("#{pip_versioned} install virtualenv psutil==5.6.6", "PYTHONDONTWRITEBYTECODE" => "1")
```
bdunne, line 49: We don't need the `PYTHONDONTWRITEBYTECODE` anymore?

Fryguy: `--no-compile` does the same thing and is what we used elsewhere: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-compile

PR #309, opened 2022-09-30
Add support for building repmgr13 v 5.2.1

jrafanie:
* Commit the contents of the enterprise db's repmgr src rpm (minus the binary)
* Make changes to this src rpm specifically to deal with changes they made that are not necessary when using dnf modules.  They assume PG 13 is installed in a separate directory than other PG versions.  Using dnf modules, you only have one enabled at a time and can rely on PG in the same location regardless of which module version you use. 
* Update our dependency in appliance to use repmgr13 >= 5.2.1
* Add Dockerfile and build script to build repmgr13

NOTE:  I've removed the ubi9/stream9 changes.

They still live in 2 additional commits [here](https://github.com/ManageIQ/manageiq-rpm_build/compare/master...jrafanie:update_repmgr_from_406_to_521_with_ubi9)
with patch files...

<details>
  <summary>More details</summary>
  <p>

```
From dcc385819ba59ebdd8b4d3c756fa9a020c7e35ca Mon Sep 17 00:00:00 2001
From: Joe Rafaniello <jrafanie@gmail.com>
Date: Wed, 19 Oct 2022 16:12:00 -0400
Subject: [PATCH 1/2] Use ubi9/stream 9

No need for postgresql modules since pg 13 is the default in stream 9

Changes to get ubi9/stream9 working:

1) pgxs.mk is needed and provided by postgresql-server-devel
2) libpgcommon.a and libpgport.a are needed and provided by postgresql-static
3) Contrib repo now contains the postgresql-server-devel and postgresql-static
packages so we need to enable crb
4) tmpfiles_create macro is either removed/deprecated and suggested to use
tmpfiles_create_package with the newer systemd in ubi9/stream9.

Note, postgresql-server-devel requires postgresql-private-devel which also
provides libpq and so conflicts with libpq-devel.  postgresql-static also
requires postgresql-server-devel.

libpgcommon.a and libpgport.a were excluded form devel packages and instead put
into the postgresql-static package and discussed in a prior commit.  See this bz
for more information:  https://bugzilla.redhat.com/show_bug.cgi?id=784281
---
 packages/repmgr13/Dockerfile  | 17 +++++++----------
 packages/repmgr13/repmgr.spec |  3 +--
 2 files changed, 8 insertions(+), 12 deletions(-)

diff --git a/packages/repmgr13/Dockerfile b/packages/repmgr13/Dockerfile
index 52c5aa6..3bf793e 100644
--- a/packages/repmgr13/Dockerfile
[...]

PR #334, opened 2022-11-07
Prefer building in mock rather than a container with rpmbuild

bdunne:
Mock ensures no network access by default and sets up a minimal environment with the BuildRequires listed in the RPM spec.

Also, we don't need to maintain another Dockerfile with its own dependencies.

Based on https://github.com/ManageIQ/manageiq-rpm_build/pull/309#discussion_r1007364482


Fryguy:
I'm fine with this, but we can't build with mock on MacOS, which is what the Dockerfile bought us.  We can use a VM however.

packages/repmgr13/README
```diff
@@ -0,0 +1,7 @@
+Build with:
+
+Fetch source files from https://rpm-manageiq-org.nyc3.digitaloceanspaces.com/index.html?prefix=sources_cache/repmgr13/ if needed.
+
+```bash
+mock --spec=./repmgr.spec -r centos-stream+epel-8-x86_64 --sources=./
```
jrafanie, line 6: How do I run this locally?

Fryguy: You will need a CentOS VM with the rpm build tools installed.

jrafanie: 🤷 I don't see the point of that.   Will it work with any centos/rhel version with any version of mock or other dependencies?  The Dockerfile provides the recipe for a repeatable build environment and gets us away from local build machines.

jrafanie: I know I'm missing the benefits (other than the sweet sweet deletions)... which is why I'm asking.  I can download the missing source files for repmgr and get rid of the hack to download the source during build if desired:

```
--undefine=_disable_source_fetch \
```

bdunne: Aside from the sweet sweet deletions 🤣...
- The build environment setup is much simpler. On RHEL/CentOS `dnf -y install epel-release` then on RHEL/CentOS/Fedora `dnf -y install mock`.
- Switching releases is also much easier since you just select a different target config `-r centos-stream+epel-9-x86_64` to switch to CentOS 9. 
- Mock found other dependency issues in the RPM spec that were not found with rpmbuild in the container.

bdunne: Also, I don't want to be in the business of maintaining a Dockerfile for every RPM that we build.

jrafanie: 🤷 I can merge but I'll probably end up resurrecting it locally when I need to do this again in several months.

jrafanie (approved):
LGTM.  I can always revert locally if I need to build it again.

Fryguy:
Backported to `oparin` in commit 4f9f75226392bacd9bca4e45fe5d6f9590f78d54.

```text
commit 4f9f75226392bacd9bca4e45fe5d6f9590f78d54
Author: Joe Rafaniello <jrafanie@users.noreply.github.com>
Date:   Thu Nov 10 14:36:31 2022 -0500

    Merge pull request #334 from bdunne/repmgr_build
    
    (cherry picked from commit ff48a2612b0bf0a2240a015f734535b3a40f5d3a)
```

PR #338, opened 2022-11-14
Script building all of the extra RPM packages

bdunne:


bin/build_extra_package.rb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+require 'manageiq/rpm_build/extra_package'
+
+Dir.chdir(ManageIQ::RPMBuild::ExtraPackage::PACKAGES_DIR) do
+  directories = (Dir["*"] - ["."]).reject { |o| !File.directory?(o) }
```
Fryguy, line 7: Minor, but for consistency, this should probably be sorted

bin/build_extra_package.rb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+require 'manageiq/rpm_build/extra_package'
+
+Dir.chdir(ManageIQ::RPMBuild::ExtraPackage::PACKAGES_DIR) do
```
Fryguy, line 6: This chdir feels unnecessary if it's only used for the Dir glob.

bin/build_extra_package.rb
bdunne: Done

bin/build_extra_package.rb
bdunne: If I don't chdir, the directories will be fully qualified paths which will require further processing.

lib/manageiq/rpm_build/extra_package.rb
```diff
@@ -0,0 +1,107 @@
...
+          "kafka"       => {
+            :mock_extras => "--enable-network",
+            :os          => "centos-stream+epel",
+          },
+          "qpid-proton" => {
+            :os => "centos-stream+epel",
+          },
+          "repmgr13"    => {
+            :os => "centos-stream+epel",
+          },
```
Fryguy, line 92: looks like you will need some extra config based on #341

PR #346, opened 2022-12-01
[WIP] Bump ansible-runner to use python3.9

Fryguy:
EPEL no longer ships ansible 5 / ansible-core 2.12, which used python 3.8, and instead ships ansible 6 / ansible-core 2.13, which uses python 3.9 instead. As such, our pre-installed runner is expecting 3.8 and installs there, but breaks when used in a more recent build.

@bdunne Please review.  master, oparin, najdorf, and morphy are all busted because of this.  The backport for najdorf and morphy will have to be different, but right now I'm focused on oparin.

WIP because I need to still fully test and verify.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,8 +52,8 @@ RUN ARCH=$(uname -m) && \
       libxslt-devel \
       make \
       openssl-devel \
-      python38-devel \
-      python38-pip \
+      python39-devel \
+      python39-pip \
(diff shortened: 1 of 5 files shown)


bdunne (approved):

bdunne:
Closed by: https://github.com/ManageIQ/manageiq-rpm_build/pull/440

PR #363, opened 2022-12-23
Fix issue where the wrong image tag was used for building RPMs

Fryguy:
This commit also moves the execution into bin/run_container_image in order to copy the logic that sets up the ENV vars that define the image tag.

@bdunne Please review.

### DIFF
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -15,7 +15,7 @@ jobs:
       run: |
         echo "REGISTRY_USERNAME=${{ secrets.DOCKER_REGISTRY_USERNAME }}" >> $GITHUB_ENV
         echo "REGISTRY_PASSWORD=${{ secrets.DOCKER_REGISTRY_PASSWORD }}" >> $GITHUB_ENV
-    - name: Run container build script
+    - name: Build and push RPM build image
       run: bin/build_container_image
   build_rpms:
     needs: build_rpm_build_container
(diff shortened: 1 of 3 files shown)


bin/build_container_image
```diff
@@ -3,7 +3,7 @@
 REGISTRY=${REGISTRY:-"docker.io"}
 ORGANIZATION=${ORGANIZATION:-"manageiq"}
 DEFAULT_TAG="latest"
-[ "$GITHUB_EVENT_NAME" == "push" -a "$GITHUB_REF_NAME" != "master" ] && \
+[[ "$GITHUB_EVENT_NAME" =~ ^(push|workflow_dispatch)$ && "$GITHUB_REF_NAME" != "master" ]] && \
```
Fryguy, line 6: I think this is part of the reason that a manual build wasn't working correctly?

Fryguy:
I'm not a fan of these build and run scripts, but for now I think it's fine - we can cleanup afterwards.

.github/workflows/ci.yaml
```diff
@@ -33,5 +33,5 @@ jobs:
       run: |
         echo "REGISTRY_USERNAME=${{ secrets.DOCKER_REGISTRY_USERNAME }}" >> $GITHUB_ENV
         echo "REGISTRY_PASSWORD=${{ secrets.DOCKER_REGISTRY_PASSWORD }}" >> $GITHUB_ENV
```
Fryguy, line 35: I'm wondering if we should drop these - that way CI isn't pushing the image to docker; only build_rpms would do that.

.github/workflows/build_rpms.yaml
```diff
@@ -26,17 +26,17 @@ jobs:
       run: |
         echo "REGISTRY_USERNAME=${{ secrets.DOCKER_REGISTRY_USERNAME }}" >> $GITHUB_ENV
         echo "REGISTRY_PASSWORD=${{ secrets.DOCKER_REGISTRY_PASSWORD }}" >> $GITHUB_ENV
-    - name: Build and push RPMs
+    - name: Prepare options file
       run: |
-        echo $REGISTRY_PASSWORD | docker login docker.io --password-stdin --username $REGISTRY_USERNAME
```
bdunne, line 31: Removing this may cause the pull to fail since it could hit the unauthenticated pull limit.

.github/workflows/build_rpms.yaml
```diff
@@ -26,17 +26,17 @@ jobs:
...
+        mkdir -p /tmp/options
         echo "---" > /tmp/options/options.yml
         echo "rpm_repository:" >> /tmp/options/options.yml
         echo "  digitalocean_access_token: ${{ secrets.RPM_BUILD_DO_ACCESS_TOKEN }}" >> /tmp/options/options.yml
         echo "  s3_api:" >> /tmp/options/options.yml
         echo "    access_key: ${{ secrets.RPM_BUILD_S3_ACCESS_KEY }}" >> /tmp/options/options.yml
         echo "    secret_key: ${{ secrets.RPM_BUILD_S3_SECRET_KEY }}" >> /tmp/options/options.yml
-        docker run -v /tmp/options:/root/OPTIONS:z docker.io/manageiq/rpm_build:latest build --build-type nightly --git-ref ${{ github.ref_name }} --update-rpm-repo
+    - name: Build and push RPMs
+      run: run_container_image
```
bdunne, line 39: Should we move the rest of the steps into that script too?

Fryguy: I was considering that, but for now I'd like to keep it this way.  Main reason is that these secrets can't be referenced from a script, and you would need lines here anyway to do something like move them into ENV vars.  In the end, we probably don't want to be modifying the yamls like this and should use ENV vars directly in the uploader script.  For now, I just wanted to fix the issue at hand.

Fryguy:
Backported to `oparin` in commit 473accce9ac814b65a4f73e40c5e33afb6de6340.

```text
commit 473accce9ac814b65a4f73e40c5e33afb6de6340
Date:   Tue Jan 3 11:09:27 2023 -0500

    Merge pull request #363 from Fryguy/fix_rpm_build_tags
    
    Fix issue where the wrong image tag was used for building RPMs
    
    (cherry picked from commit 271395184220ed5f90e4f0e0d1da34b5126389d7)
```

PR #386, opened 2023-03-27
Reduce memory usage of nightly build purger

bdunne:
Previously I saw memory usage over 17GB.


Fryguy:
I can't understand how this drops the memory usage from 17gb, but the code looks good, so merging.

bdunne:
The arrays being allocated are much smaller since there aren't a bunch of `nil`s at the beginning.

PR #399, opened 2023-06-27
Revert "Merge pull request #393 from bdunne/revert_278"

bdunne:
This reverts commit d27d2219c1a6a75e96519697800dcab057b93581, reversing changes made to 124c2dd57710780302440f481d10b316e54a6b4c.

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -60,7 +60,8 @@ module ManageIQ
           shell_cmd("yarn install")
           shell_cmd("yarn run available-languages")
           shell_cmd("yarn run build")
-          shell_cmd("git clean -xdf") unless RUBY_PLATFORM.include?("s390x") # git clean would unset the yarn version change
+          shell_cmd("git clean -xdf")  # cleanup temp files
+          shell_cmd("yarn set version 1.22.18") if RUBY_PLATFORM.include?("s390x")
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -60,7 +60,7 @@ def build_service_ui
           shell_cmd("yarn install")
           shell_cmd("yarn run available-languages")
           shell_cmd("yarn run build")
-          shell_cmd("git clean -xdf")  # cleanup temp files
+          shell_cmd("git clean -xdf") unless RUBY_PLATFORM.include?("s390x") # git clean would unset the yarn version change
```
bdunne, line 63: Any concern that we'll ship extra temp files here? I wonder if we should `git clean` on s390x then `yarn set version` again.

bdunne: Something like this maybe?
```suggestion
          shell_cmd("git clean -xdf") # cleanup temp files
          shell_cmd("yarn set version 1.22.18") if RUBY_PLATFORM.include?("s390x")
```

kbrock: Is this running in the target container, or on the host that is then copied to the container?

Think when we run stuff like this inside the container, it makes the image disk size larger.

Fryguy: > Any concern that we'll ship extra temp files here? I wonder if we should git clean on s390x then yarn set version again.

No, the git clean is to ensure that license finder is working with the original source.  I think that the yarn set version after the git clean is acceptable.

Fryguy: I think I misspoke a bit - this was partly for license_finder reasons in case anything brought in any "extra" files.  But this was more for payload size and removing unneeded build artifacts.  The git clean has been there since Initial commit.

So yeah, @bdunne I like your approach better.

PR #400, opened 2023-06-28
[MORPHY] Lockdown ansible to 5.4.0-3

Fryguy:
Lockdown ansible to 5.4.0-3

(cherry picked from commit 82b319923a3ea1acb75f5f918b86a5900bb565d5)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,6 @@ RUN dnf -y remove subscription-manager* && \
     dnf -y group install "development tools" && \
     dnf config-manager --setopt=epel.exclude=*qpid-proton* --setopt=tsflags=nodocs --save && \
     dnf -y install \
-      ansible \
       cmake \
       copr-cli \
       createrepo \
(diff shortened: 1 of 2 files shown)


rpm_spec/subpackages/manageiq-core
```diff
@@ -13,6 +13,7 @@ Requires: git
 Requires: net-snmp
 Requires: net-snmp-libs
 Requires: net-snmp-utils
+Requires: python38
```
Fryguy, line 16: Not sure about this one, but I can't see why this would cause freezing

Dockerfile
```diff
@@ -42,7 +41,9 @@ RUN dnf -y remove subscription-manager* && \
       postgresql-server-devel \
       qpid-proton-c-devel \
       ruby-devel \
-      wget && \
+      wget \
+      # For seeding ansible runner with ansible-galaxy
+      ansible-5.4.0-3.el8 && \
```
bdunne, line 46: Any reason to add this to the end instead of editing the deleted line above?

Fryguy: It's mainly because it had the comment attached and I didn't want it to spray that all of it was just for ansible.  Also this was a cherry pick and I did the same thing in the original PR

PR #408, opened 2023-07-28
create an alias on upgrade

kbrock:
related: https://github.com/ManageIQ/manageiq/pull/22392

Before
======

If evmserverd is enabled on a previous version of the product, then an alias is not created. Upgrading will not create the alias.

Enabling after the upgrade will create the alias (symbolic link)

After
=====

Upgrading the rpm manually creates the alias / symbolic link in the filesystem. If the service was previously not enabled, the symbolic link was unnecessarily created.

There is a bunch of logic / PR fixes in systemd around aliases already existing or being invalid so this edge case seems like it will be fine


kbrock:
Not sure the best way to setup an old appliance and upgrading these rpms on an m1 mac

rpm_spec/subpackages/manageiq-appliance
```diff
@@ -68,6 +68,15 @@ Requires: cockpit-ws
 #motd is owned by system
 %{__cp} -f %{_sysconfdir}/motd.manageiq %{_sysconfdir}/motd
 
+# run this if we are upgrading
```
Fryguy, line 71: Can this comment be more descriptive on what this code is doing?

Fryguy:
I'm not sure about this since I'm not sure it's really needed - this is mostly a dev convenience at the moment, right?  Perhaps put another way, can this code go away when we fully get to the manageiq.service being the official name, or will this linking code stick around?

rpm_spec/subpackages/manageiq-appliance
kbrock: ok, that was describing the gt 1 bit
The document for what is happening inside is described below.

I'll try and do a better job writing the documentation for the command

kbrock:
between documentation and cross service links (across repos), this is a big effort.
The attempt to do it all at once got too much.

So just introducing the alias seemed like a good idea for now.

If we can have the user disable and then re-enable evmserverd, then this code is not necessary.

But short of that, we will need code like this whether we rename or alias the service.

kbrock:
update:
- rebased
- changed git comments
- added _unitdir macro

/cc @bdunne does this unitdir look right to you?
(this is low priority)

agrare:
:+1: for the `__unitdir`
https://rpm-packaging-guide.github.io/ says `_unitdir   /usr/lib/systemd/system`
We have `lib/systemd/system` but `/lib -> usr/lib` so this is the same thing.

Fryguy:
Seems good to me - I'll leave it up to @agrare

agrare:
I'm going to build this locally and test an upgrade on my appliance, but LGTM

agrare:
Having to upgrade my centos-8-stream to 9 and that is giving me some issues.

agrare:
Okay thanks to @bdunne I was able to test this out on a stream 9 appliance

```
# dnf install --nobest /tmp/rpms/*.rpm
...
[root@manageiq ~]# ls -l /lib/systemd/system/manageiq.service 
lrwxrwxrwx. 1 root root 42 Feb 27 09:54 /lib/systemd/system/manageiq.service -> /usr/lib/systemd/system/evmserverd.service
[root@manageiq ~]# systemctl status manageiq.service
○ evmserverd.service - EVM server daemon
     Loaded: loaded (/usr/lib/systemd/system/evmserverd.service; enabled; preset: disabled)
     Active: inactive (dead)
```

So with an already enabled evmserverd.service the rpm upgrade created the alias symlink properly and systemctl picked it up immediately.

LGTM

agrare (approved):

PR #421, opened 2023-09-21
No longer need to build postgres ourselves.

bdunne:
10.23 is available upstream


Fryguy:
Do we need to backport this or do old release still build from master?

PR #423, opened 2023-10-27
[QUINTEROS] Bump rpm version to 17.1.0

Fryguy:
@bdunne Please review

### DIFF
--- a/config/options.yml
+++ b/config/options.yml
@@ -14,7 +14,7 @@ repos:
 version:
 release:
 rpm:
-  version: 17.0.0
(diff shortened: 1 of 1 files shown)


bdunne:
Looks like the conflict is from the merge of master into quinteros https://github.com/ManageIQ/manageiq-rpm_build/blob/64bc4a07bc250c9e59eff85e2887b6577abd8b71/config/options.yml#L17

Fryguy:
@bdunne I created #424 instead and included this commit in it.

Issue #411, opened 2023-08-11
Disable rsyslog/journald log copies on appliances


Fryguy:
Opening this issue to discuss.  The way to implement this is to remove im_journal module lines from /etc/rsyslog.conf, however since we don't own that file, it will require a post script in the appliance RPM.  Is there a downside to doing this?

@agrare @bdunne  Thoughts?

agrare:
Okay had to dust off my unix handbook to get this done :laughing: 
https://github.com/ManageIQ/manageiq-rpm_build/pull/426 will comment out all lines starting with `module(load="imjournal"` and ending with `StateFile="imjournal.state")`

PR #426, opened 2023-11-13
Disable rsyslog duplication of systemd-journal

agrare:
By default systemd-journal is non-persistent and rsyslog is configured to copy everything from the journal to `/var/log/syslog`.

If you have a `/var/log/journal/` directory on persistent storage then systemd-journal will automatically persist journal files but rsyslog doesn't also automatically disable copying to syslog

TODO:
- [x] Live test of an rpm upgrade

Fixes https://github.com/ManageIQ/manageiq-rpm_build/issues/411

### DIFF
--- a/rpm_spec/subpackages/manageiq-appliance
+++ b/rpm_spec/subpackages/manageiq-appliance
@@ -100,6 +100,9 @@ fi
 # Disable rsyslog duplicating systemd-journal output
 # This will comment out the multi-line module load from /etc/rsyslog.conf
 sed -i '/module(load="imjournal"/, /StateFile="imjournal.state")/ s|^|#|' %{_sysconfdir}/rsyslog.conf
+if systemctl is-active --quiet rsyslog; then
+  systemctl restart rsyslog
+fi
 
 %files appliance
 %defattr(-,root,root,-)
--- a/rpm_spec/subpackages/manageiq-appliance
+++ b/rpm_spec/subpackages/manageiq-appliance
@@ -99,7 +99,7 @@ fi
 
 # Disable rsyslog duplicating systemd-journal output
 # This will comment out the multi-line module load from /etc/rsyslog.conf
-sed -i '/^module(load="imjournal"/, /StateFile="imjournal.state")/ s|^|#|' %{_sysconfdir}/rsyslog.conf
+sed -i '/^module(load="imjournal"/, /^\s+StateFile="imjournal.state")/ s|^|#|' %{_sysconfdir}/rsyslog.conf
 if systemctl is-active --quiet rsyslog; then
   systemctl restart rsyslog
 fi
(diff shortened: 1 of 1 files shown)


Fryguy:
@bdunne Please review.

rpm_spec/subpackages/manageiq-appliance
```diff
@@ -82,6 +82,10 @@ if systemctl is-active --quiet evm-failover-monitor; then
   systemctl restart evm-failover-monitor
 fi
 
+# Disable rsyslog duplicating systemd-journal output
+# This will comment out the multi-line module load from /etc/rsyslog.conf
+sed -i '/module(load="imjournal"/, /StateFile="imjournal.state")/ s|^|#|' %{_sysconfdir}/rsyslog.conf
+
```
agrare, line 88: TODO should probably restart rsyslog.service after this config change

agrare:
Okay with the `sed` scriptlet here this is what rsyslog.conf looks like after `dnf install`

```
#### MODULES ####

module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
       SysSock.Use="off") # Turn off message reception via local log socket; 
                          # local messages are retrieved through imjournal now.
#module(load="imjournal"            # provides access to the systemd journal
#       UsePid="system" # PID nummber is retrieved as the ID of the process the journal entry originates from
#       StateFile="imjournal.state") # File to store the position in the journal
#module(load="imklog") # reads kernel messages (the same are read from journald)
#module(load="immark") # provides --MARK-- message capability

# Provides UDP syslog reception
# for parameters see http://www.rsyslog.com/doc/imudp.html
#module(load="imudp") # needs to be done just once
#input(type="imudp" port="514")
```

agrare:
Okay retested after adding the `systemctl restart rsyslog.service` line

rpm_spec/subpackages/manageiq-appliance
```diff
@@ -82,6 +82,13 @@ if systemctl is-active --quiet evm-failover-monitor; then
   systemctl restart evm-failover-monitor
 fi
 
+# Disable rsyslog duplicating systemd-journal output
+# This will comment out the multi-line module load from /etc/rsyslog.conf
+sed -i '/^module(load="imjournal"/, /StateFile="imjournal.state")/ s|^|#|' %{_sysconfdir}/rsyslog.conf
```
agrare, line 87: If you aren't a `sed` expert (which I am not), this means "starting at this regex" (`/^module(load="imjournal"/`) until this regex (`/StateFile="imjournal.state")/`), substitute the start of the line with `#` (`s|^|#|`)

Fryguy:
> And confirmed that after restarting rsyslog.service journal messages are no longer duplicated to /var/log/messages

Just to clarify. I thought this would prevent them from being written to `/var/log/journal` ?  Only reason I'm asking is that /var/log/messages is in the settings as one of the places we collect logs from, so will this affect log collection?

agrare:
@Fryguy we also pull logs from `/var/log/journal/` and we already use `journalctl` on uploaded log bundles

Fryguy (approved):
LGTM - @bdunne Please also review.

agrare:
Hey @bdunne could you take a look at this?

kbrock:
This looks great. anything special to help get it in?

PR #429, opened 2023-12-18
Ruby 3.1

jrafanie:
Start the switch to ruby 3.1.

* use ruby 3.1 with explicit minimum of 0.37.0 qpid_proton from epel8/9
* drop qpid_proton sources
* use yaml unsafe load when reading the sti loader yaml as we control the file and it intentionally contains all of our code constants.

Part of https://github.com/ManageIQ/manageiq/issues/22696

I was able to get this working on mac with colima:
* `bin/build_container_image`
*  `docker run -v /tmp/options:/root/OPTIONS:z manageiq/rpm_build:latest build --build-type nightly`

I modified the options.yml locally to build and use this PR via the options.yml:
https://github.com/ManageIQ/manageiq/pull/22808.  That PR temporarily included the qpid proton bump from https://github.com/ManageIQ/manageiq/pull/22271.


lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -104,7 +104,7 @@ def plugin_paths
 
       def fixup_sti_loader!
         sti_loader_yml_path = miq_dir.join("tmp/cache/sti_loader.yml")
-        sti_loader          = YAML.load_file(sti_loader_yml_path)
+        sti_loader          = YAML.unsafe_load(File.read(sti_loader_yml_path))
```
jrafanie, line 107: note, `unsafe_load_file` didn't exist in all versions of psych so I use the more verbose but backward compatible code.

config/options.yml
```diff
@@ -49,7 +49,6 @@ rpm_repository:
         :python-pylxca: !ruby/regexp /.+-2\.1\.1.+/
         :python-unittest2: !ruby/regexp /.+-1\.1\.0.+/
         :python-vspk: !ruby/regexp /.+-5\.3\.2.+/
-        :qpid-proton: !ruby/regexp /.+-0\.30\.0.+/
```
Fryguy, line 52: do we intend to backport this to quinteros?  (I don't think we wanted to, but just checking)

rpm_spec/subpackages/manageiq-gemset
```diff
@@ -30,7 +30,7 @@ Requires: python3-pylxca
 Requires: cyrus-sasl
 Requires: cyrus-sasl-plain
 Requires: python3-vspk
-Requires: qpid-proton-c
+Requires: qpid-proton-c >=  0.37.0
```
Fryguy, line 33:
```suggestion
Requires: qpid-proton-c >= 0.37.0
```

config/options.yml
jrafanie: We haven't decided if we're shipping this 3.1 now or later.  It's still TBD.

PR #435, opened 2024-01-02
Add manageiq user to disk group

nasark:
Since we run as non-root, this will allow smartstate scans to be run on ISCSI store types

@miq-bot assign @agrare 
@miq-bot add_label bug, quinteros/yes?

### DIFF
--- a/rpm_spec/subpackages/manageiq-appliance
+++ b/rpm_spec/subpackages/manageiq-appliance
@@ -65,6 +65,8 @@ Requires: cockpit-ws
 # Add sub uids/gids for running non-privileged containers
 grep manageiq /etc/subuid >/dev/null 2>&1 || /usr/sbin/usermod --add-subuids 100000-165535 manageiq
 grep manageiq /etc/subgid >/dev/null 2>&1 || /usr/sbin/usermod --add-subgids 100000-165535 manageiq
+# Add manageiq user to disk group to allow ISCSI smartstate scans
+groups manageiq | grep disk >/dev/null 2>&1 || /usr/sbin/usermod --append --groups disk manageiq
 
 %posttrans appliance
 %{_bindir}/systemctl try-restart evmserverd.service


rpm_spec/subpackages/manageiq-appliance
```diff
@@ -65,6 +65,7 @@ Requires: cockpit-ws
 # Add sub uids/gids for running non-privileged containers
 grep manageiq /etc/subuid >/dev/null 2>&1 || /usr/sbin/usermod --add-subuids 100000-165535 manageiq
 grep manageiq /etc/subgid >/dev/null 2>&1 || /usr/sbin/usermod --add-subgids 100000-165535 manageiq
+/usr/sbin/usermod -a -G disk manageiq
```
Fryguy, line 68: Can you add a comment here? Otherwise it looks like it's part of the previous comments.

Fryguy, line 68: Also, very minor, but for future readers, I wonder if it's easier to use the full option names?

```suggestion
/usr/sbin/usermod --append --groups disk manageiq
```

agrare: We should check what the command does when the user is already part of the group, if it returns a non-zero exit code that will cause the rpm scriptlet to fail

agrare:
```
[root@manageiq ~]# groups manageiq
manageiq : manageiq
[root@manageiq ~]# /usr/sbin/usermod --append --groups disk manageiq
[root@manageiq ~]# echo $?
0
[root@manageiq ~]# groups manageiq
manageiq : manageiq disk
[root@manageiq ~]# /usr/sbin/usermod --append --groups disk manageiq
[root@manageiq ~]# echo $?
0
```

Seems like it returns `0` in either case, might be a good idea to check if it is in the group first though `groups manageiq | grep disk >/dev/null 2>&1 || /usr/sbin/usermod ...` just so it is more obvious

nasark:
Ready for another pass

Fryguy:
Backported to `quinteros` in commit acab9a30c34e093f55ad89bbebd1c0631a7d8eeb.

```text
commit acab9a30c34e093f55ad89bbebd1c0631a7d8eeb
Date:   Tue Jan 9 15:50:13 2024 -0500

    Merge pull request #435 from nasark/add_miq_user_to_disk_group
    
    Add manageiq user to disk group
    
    (cherry picked from commit 658af63fc1f5da6cda4e827d88572ad25a7bd7ca)
```

PR #437, opened 2024-01-10
Faster rpm build step 1

jrafanie:
* Don't create .build-id artifacts
* Use built-in vcs exclusion for tar to also avoid .gitignore and other vcs files
* Curl and use the rpm 4.19.1 version of brp-strip for the xargs parallel processing and filtering - add our filtering patch for ruby and js on top of it.

Extracted from https://github.com/ManageIQ/manageiq-rpm_build/pull/436

generate_rpm is about 57 minutes, total build is 1 hour 15 minutes, roughly 30 minutes faster with these changes, of which nearly all of it is the brp-strip.

EDIT:  The rpm metadata and contents looks to be just what we want changed.  I've posted an update with screenshots from the `pkgdiff -details -hide-unchanged beforeRPM afterRPM` below.

### DIFF
--- a/container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
+++ b/container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
@@ -1,0 +1,33 @@
+From 7d0f0d9c34a934327cfdd68e721c250af39a5c13 Mon Sep 17 00:00:00 2001
+From: Joe Rafaniello <jrafanie@gmail.com>
+Date: Tue, 9 Jan 2024 15:00:25 -0500
+Subject: [PATCH] Add js/rb filtering on top of 4.19.1
+
+---
+ scripts/brp-strip | 3 +++
+ 1 file changed, 3 insertions(+)
+
+diff --git a/scripts/brp-strip b/scripts/brp-strip
+index 799bf2bc2..b258debc8 100755
+--- a/scripts/brp-strip
++++ b/scripts/brp-strip
+@@ -20,6 +20,7 @@ esac
+ 
+ # Below is the explanation of commands in the order of their appearance
+ # Ignore /usr/lib/debug entries
++# Ignore all js and rb javascript and ruby files
+ # Ignore all go(guile objects & golang) files
+ # Consider files with only single link
+ # Run the file command to find relevant non-stripped binaries, with bundle size of 32
+@@ -33,6 +34,8 @@ strip_elf_binaries()
+ 
+   find "$RPM_BUILD_ROOT" -type f \
+     ! -regex "${RPM_BUILD_ROOT}/*usr/lib/debug.*" \
++    ! -name "*.js" \
++    ! -name "*.rb" \
+     ! -name "*.go" -links "${nlinks}" -print0 | \
+     xargs -0 -r -P${nprocs} -n${MAX_ARGS} sh -c "file \"\$@\" | \
+     sed -n -e 's/^\(.*\):[ 	]*ELF.*, not stripped.*/\1/p' | \
+-- 
+2.42.0
+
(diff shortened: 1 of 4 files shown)


Fryguy (approved):

Fryguy:
Awesome work @jrafanie !

LGTM

Fryguy:
@jrafanie what is left to do here / why is this WIP?

container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
```diff
@@ -0,0 +1,33 @@
...
++# Ignore all js and rb javascript and ruby files
+ # Ignore all go(guile objects & golang) files
+ # Consider files with only single link
+ # Run the file command to find relevant non-stripped binaries, with bundle size of 32
+@@ -33,6 +34,8 @@ strip_elf_binaries()
+ 
+   find "$RPM_BUILD_ROOT" -type f \
+     ! -regex "${RPM_BUILD_ROOT}/*usr/lib/debug.*" \
++    ! -name "*.js" \
++    ! -name "*.rb" \
```
Fryguy, line 27: For the future I wonder if we can upstream a change where we can specify a list of file extensions to ignore?  Then maybe we don't have to carry a patch.

jrafanie:
Looks like the python compiled programs changed slightly, which could be due to changes since they were run a different times or due to the strip changes.  I guess we'll only know when we try them.  

image

The vcs ignore files are removed as expected.

image

The venv provides sections looks to be unchanged, as desired... we'll tackle the pythondeps slowness elsewhere:

image

the build-id metadata is no longer packaged up:
image

jrafanie:
ok @bdunne @Fryguy this is ready to go.  I don't know how to test the bstrip change.  Technically, the 4.19 version brought no new dependencies into it so it should be fairly safe but clearly, it's changed so we should see if we can smoke test them.

container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
jrafanie: It would be nice if find accepted a pattern file exclusion list so you don't have to deal with escaping things.

jrafanie:
@Fryguy let me know if there's anything else needed for this PR.  I think it's ready to go and is limited to what we thought was safe to go with.

jrafanie:
cc @bdunne (same as above... throw 🍅 )

Fryguy (approved):
LGTM - just want @bdunne as a second set of eyes

container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
kbrock: I pushed this change upstream. lets see if it sticks
https://github.com/rpm-software-management/rpm/pull/2858

Fryguy:
Backported to `quinteros` in commit ed6f1e80e10631620dbfb284b81a81ae9b35e6ed.

```text
commit ed6f1e80e10631620dbfb284b81a81ae9b35e6ed
Date:   Thu Jan 18 14:37:21 2024 -0500

    Merge pull request #437 from jrafanie/faster_rpm_build_step1
    
    Faster rpm build step 1
    
    (cherry picked from commit 001b88f11e4e66ab1ef43cfa9e6a9f30b4f24257)
```

container-assets/Add-js-rb-filtering-on-top-of-4.19.1.patch
Fryguy: @kbrock @jrafanie Your PR was included in `rpm-4.20.0-alpha`

```
$ git tag --contains  d99d625
rpm-4.20.0-alpha
```

PR #440, opened 2024-01-25
Upgrade to UBI9

bdunne:
Includes upgrading python venv for python3.9 (minimum for EL9)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi
+FROM registry.access.redhat.com/ubi9/ubi
 
 ENV TERM=xterm \
     APPLIANCE=true \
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -52,8 +52,8 @@ RUN ARCH=$(uname -m) && \
       libxslt-devel \
       make \
       openssl-devel \
-      python38-devel \
-      python38-pip \
+      python3-devel \
+      python3-pip \
```
Fryguy, line 56: Is it possible to do python39-devel and python39-pip?  Reason is that while it may be the system python, moving forward we're going to want to go to 3.10, 3.11, etc, and it's so much easier if we are explicit on the version.

rpm_spec/subpackages/manageiq-ansible-venv
```diff
@@ -4,7 +4,7 @@ Summary: %{product_summary} Ansible Runner Virtual Environment
 %global __requires_exclude ^\/var\/lib\/manageiq\/venv\/bin\/python
 
 Requires: ansible >= 5, ansible < 6
-Requires: python38
+Requires: python39
```
Fryguy, line 7: I'm not sure if this works or not - I thought ansible 5 required 3.8, but if you can get a build I can verify it.

bdunne: I'm not sure either.  I see this in the ansible RPM spec:
```
%if 0%{?rhel} == 8
# RHEL 8's ansible-core package is built using Python 3.8, which is not the default version.
%define python3_pkgversion 38
BuildRequires:  python%{python3_pkgversion}-rpm-macros
%endif
```
So, it may be using system python in EL9.

bdunne: I will update that and rebuild it based on your comment above

Dockerfile
bdunne: No, python 3.9 RPMs are python3-*.  There is no python3.10.  Python 3.11 is `python-3.11` or `python-311`, but not all of the dependencies that we need are there yet.

rpm_spec/subpackages/manageiq-gemset
```diff
@@ -34,7 +30,6 @@ Requires: qpid-proton-c >= 0.37.0
 
 # For Appliance Console
 Requires: libsodium-devel
-Requires: network-scripts
```
bdunne, line 37: TODO 😅

lib/manageiq/rpm_build/generate_ansible_venv.rb
```diff
@@ -46,7 +46,12 @@ def populate
       def install_python_system_packages
         where_am_i
 
-        shell_cmd("#{pip_versioned} install --no-compile virtualenv ansible-runner")
+        shell_cmd("#{pip_versioned} install --no-compile virtualenv")
```
bdunne, line 49: this can be dnf installed (may need pip3 also, but just for build)

lib/manageiq/rpm_build/generate_ansible_venv.rb
```diff
@@ -86,10 +92,10 @@ def move_content_to_build_dir
         # Copy the venv dir
         FileUtils.mv(@venv_build_dir, venv_dir)
 
-        # Copy python 3.8 site-packages for ansible-runner
+        # Copy python site-packages for ansible-runner
         site_packages_dir = venv_dir.join("site-packages")
         FileUtils.mkdir_p(site_packages_dir)
-        dirs_to_copy = Dir.glob("/usr/local/lib/python3.8/site-packages/*").grep_v(%r{/(distlib|filelock|platformdirs|virtualenv)})
+        dirs_to_copy = Dir.glob("/usr/local/lib/python#{VENV_PYTHON_VERSION}/site-packages/*").grep_v(%r{/(distlib|filelock|platformdirs|virtualenv)})
```
bdunne, line 98: We don't need to grep -v if we're using RPMs for virtualenv

lib/manageiq/rpm_build/generate_ansible_venv.rb
```diff
@@ -46,7 +46,12 @@ def populate
       def install_python_system_packages
         where_am_i
 
-        shell_cmd("#{pip_versioned} install --no-compile virtualenv ansible-runner")
+        shell_cmd("#{pip_versioned} install --no-compile virtualenv")
+
+        # Install ansible-core dependencies which aren't auto-installed with ansible-core
+        shell_cmd("#{pip_versioned} install --no-compile 'jinja2>=3.0.0' 'resolvelib<0.9.0,>=0.5.3' packaging")
```
bdunne, line 52: Doesn't look like we need these.  packagind and resolvelib are already installed as system packages and jinja doesn't appear to be needed

Dockerfile
```diff
@@ -10,19 +10,19 @@ RUN chmod +t /tmp
...
+      https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm \
+      https://rpm.manageiq.org/release/18-radjabov/el9/noarch/manageiq-release-18.0-1.el9.noarch.rpm && \
     dnf -y module enable ruby:3.1 && \
     dnf -y module enable nodejs:18 && \
-    dnf -y module disable virt:rhel && \
-    if [ ${ARCH} != "s390x" ] ; then dnf config-manager --setopt=ubi-8-*.exclude=rpm* --save; fi && \
     dnf -y group install "development tools" && \
     dnf config-manager --setopt=tsflags=nodocs --save && \
-    dnf -y install \
+dnf -y install \
```
Fryguy, line 25: whitespace accidentally removed?

bdunne: Fixed

PR #442, opened 2024-02-14
Drop WMI preparing for EL9

bdunne:
Related to: https://github.com/ManageIQ/manageiq/pull/22279


Fryguy:
I don't think we ever cut a new manageiq-smartstate from master with the winrm stuff removed?  Or at least I don't see a tag.  @agrare Do you know?

agrare:
> Or perhaps wmi isn't actually needed for smartstate? It's hard to tell.

I think it was, https://github.com/ManageIQ/manageiq-smartstate/pull/172/files#diff-f6375ed0ad0afbc523e281b5f41a89995acb6c2b6dd73ca512baf8684121d2beL45

And no I don't see a gem release including the winrm removal either.

PR #450, opened 2024-03-14
Add opentofu-runner service

agrare:
Depends on:
- [ ] https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/2

### DIFF
--- a/rpm_spec/subpackages/manageiq-gemset-services
+++ b/rpm_spec/subpackages/manageiq-gemset-services
@@ -32,3 +32,4 @@ done
 
 %files gemset-services
 %{_prefix}/lib/systemd/system/manageiq-providers*
+%{_prefix}/lib/systemd/system/opentofu-runner*


Fryguy:
Can you add this type of thing to the pluggability checklist as well?

agrare:
I can add it but I'm not sure how pluggable the RPM build is or can be, not aware of any other aspect of it that we have overridden in plugins @bdunne ?

Fryguy:
We have systemd units pluggable but it's pluggable generically by name (i.e. manageiq-providers-*) on the rpm side... Not sure how we can do it any differently

agrare:
Right the opentofu-runner.service content is already defined in the provider repo, one thing that comes to mind is if we don't care about splitting up "core-services" and "gemset-services" we could just take `%{_prefix}/lib/systemd/system/*.{service,target}`

Fryguy:
Yeah I'm not sure why we have them split other than the rpms themselves are split that way

PR #471, opened 2024-06-20
Fix RPM Build unpackaged file error

bdunne:
Installed (but unpackaged) file(s) found:
   /etc/sudoers.d/lvm

Introduced in: https://github.com/ManageIQ/manageiq-appliance/pull/387


Fryguy:
cc @nasark

nasark:
Thanks @bdunne

PR #478, opened 2024-07-17
Fix service ui productization symlinking

Fryguy:
The problem was that symlink_plugin_paths tries to symlink the client directory itself, but that directory already exists, so the ln_s call fails.  However, it doesn't raise an exception and just moves on.  I can't reuse that method without seriously rewriting it, so I wrote something custom.  I do plan on writing something a bit more robust, but this should fix the problem at hand.

@bdunne Please review.
cc @GilbertCherrie

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -53,7 +53,7 @@ module ManageIQ
       end
 
       def build_service_ui
-        symlink_plugin_paths("manageiq-ui-service", ui_service_dir)
+        symlink_service_ui_productization
 
         Dir.chdir(ui_service_dir) do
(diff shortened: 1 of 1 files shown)


Fryguy:
Note I did not test this with a productization plugin, because I'm not sure how.  We missed this because it doesn't happen in the manageiq build, and only happens is downstream builds.

Fryguy:
Backported to `radjabov` in commit 42cebeeae63ce7499b46cd280b43c64c7ec7b517.

```text
commit 42cebeeae63ce7499b46cd280b43c64c7ec7b517
Date:   Tue Sep 3 17:00:33 2024 -0400

    Merge pull request #478 from Fryguy/fix_service_ui_productization
    
    Fix service ui productization symlinking
    
    (cherry picked from commit 413dfbe20ef7529ca383a52fdf14e57eb22a17f4)
```

PR #484, opened 2024-08-21
\s+ wasn't matching multiple spaces and failed to match the last line commenting out the rest of the file.

bdunne:
The + needs to be escaped for it to match the last line

Example failure:
```
× rsyslog.service - System Logging Service
     Loaded: loaded (/usr/lib/systemd/system/rsyslog.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Wed 2024-08-21 09:45:27 EDT; 18min ago
       Docs: man:rsyslogd(8)
             https://www.rsyslog.com/doc/
    Process: 37598 ExecStart=/usr/sbin/rsyslogd -n $SYSLOGD_OPTIONS (code=exited, status=1/FAILURE)
   Main PID: 37598 (code=exited, status=1/FAILURE)
        CPU: 69ms

Aug 21 09:45:27 mmim421.example.com systemd[1]: rsyslog.service: Scheduled restart job, restart counter is at 5.
Aug 21 09:45:27 mmim421.example.com systemd[1]: Stopped System Logging Service.
Aug 21 09:45:27 mmim421.example.com systemd[1]: rsyslog.service: Start request repeated too quickly.
Aug 21 09:45:27 mmim421.example.com systemd[1]: rsyslog.service: Failed with result 'exit-code'.
Aug 21 09:45:27 mmim421.example.com systemd[1]: Failed to start System Logging Service.
```

PR #485, opened 2024-08-22
Bring back syslog lines that were incorrectly disabled

kbrock:
Followup to https://github.com/ManageIQ/manageiq-rpm_build/pull/484

We commented out imjournal via sed when installing the appliance. Unfortunately, instead of commenting out just the imjournal, we commented out to the end of the file.

This script removes all those extra comments to the end of the file.

There is a chance that we will incorrectly remove comments.
This will happen if the user has `##module(load="imklog")`
If support general appliances, then we may have to factor this unlikely scenario


Fryguy:
Duplicate of #486

PR #482, opened 2024-08-20
[quinteros] Allow timestamped release builds

bdunne:
- Allow timestamped RPMs in the release repo if they're in the release dir
- Ensure the prefix ends with a /
    This ensures that nightly builds from the manageiq-nightly directory are not included in the release repo when searching for release builds.
    "builds/manageiq/" vs. "builds/manageiq" which would include "builds/manageiq-nightly"

Fixes: https://github.com/orgs/ManageIQ/discussions/23135
Needs to be forward-ported to radjabov and master

PR #489, opened 2024-08-26
Allow timestamped release builds

bdunne:
Forward-port of #482


Fryguy:
radjabov in is #490 
quinteros is in #482

PR #493, opened 2024-09-05
Python requirements update

kbrock:
Overview
========

Our current python packages are not compatible with the current ansible version installed (and/or the current python version).

CP4AIOPS-5619 / https://github.com/ManageIQ/manageiq-rpm_build/pull/495 was just focused on vmware and aws.

This provides all requirements.

Issues
======

In a container, not able to build gssapi (for pyspnego[kerberos], for ms/AD)
This downloads many versions of `gssapi` but keeps failing.
This may be because I'm doing this inside docker instead of at install time.

Reference
=====

Dependencies brought up to date for:

- amazon/aws
- ansible/netcommon
- ansible/utils
- awx/awx
- azure/azcollection
- cisco/aci
- cisco/asa
- cisco/dnac
- cisco/intersight
- cisco/ios
- cisco/iosxr
- cisco/ise
- cisco/mso
- cisco/nxos
- cisco/ucs
- community/aws
- community/okd
- community/vmware
- google/cloud
- kubernetes/core
- microsoft/ad
- openstack/cloud
- ovirt/ovirt
- theforeman/foreman
- vmware/vmware_rest

### DIFF
--- a/bin/requirements.rb
+++ b/bin/requirements.rb
@@ -1,0 +1,207 @@
+#!/usr/bin/env ruby
+
+# This script takes the existing requirements.txt file
+# and updates it with the version for our supported packages
+#
+# USAGE:
+# 1. Setup environment
+#
+#     source /var/lib/manageiq/venv/bin/activate
+#     upload bin/requirements.rb and config/requirements.txt to /tmp
+#     chmod 755 requirements.rb
+#
+# 1. Get all module requirements (don't include documentation or testing ones)
+#
+#     ./requirements.rb ./requirements.txt /usr/lib/python3.9/site-packages/ansible_collections/ > new_requirements.txt
+#
+# 2. Resolve conflicts and determine if new one is correct
+#
+#     diff {,new_}requirements.txt
+#     # cp new_requirements.txt requirements.txt
+#
+
+class ParseRequirements
+  PACKAGES=%w[
+    amazon/aws/requirements.txt
+    ansible/netcommon/requirements.txt
+    ansible/utils/requirements.txt
+    awx/awx/requirements.txt
+    azure/azcollection/requirements-azure.txt
+    cisco/intersight/requirements.txt
+    community/aws/requirements.txt
+    community/okd/requirements.txt
+    community/vmware/requirements.txt
+    google/cloud/requirements.txt
+    kubernetes/core/requirements.txt
+    openstack/cloud/requirements.txt
+    ovirt/ovirt/requirements.txt
+    theforeman/foreman/requirements.txt
+    vmware/vmware_rest/requirements.txt
+  ]
+  attr_reader :filenames, :non_modules, :final
+
+  def initialize
+    @filenames = []
+    @non_modules = []
+
+    @final = {}
+  end
+
+  def add_target(filename)
+    if Dir.exist?(filename)
+      add_dir(filename)
+    elsif File.exist?(filename)
+      add_file(filename)
+    else
+      $stderr.puts("File not found: #{filename}")
+    end
+  end
+
+  def add_file(filename)
+    @filenames << filename
+    @non_modules << filename unless filename.include?("ansible_collections")
+
+    self
+  end
+
+  def add_dir(dirname)
+    dirname = dirname[0..-2] if dirname.end_with?("/")
+    PACKAGES.each do |package|
+      filename = "#{dirname}/#{package}"
+      if File.exist?(filename)
+        @filenames << filename
+      else
+        $stderr.puts("NOTICE: missing #{filename}")
+      end
+    end
+
+    self
+  end
+
+  def parse
+    filenames.each do |fn|
+      # the list of requirements-files can have items commented out - ignore those
+
+      mod = module_name_from_filename(fn)
+      IO.foreach(fn, chomp: true).each do |line|
+        lib, ver = parse_line(line)
+        next unless lib
+
+        # skip git libraries. git>= line from vsphere gave us problems
+        next if lib.start_with?("git")
+
+        # Do not version packages that are provided by the system: requests, requests[security], pyspnego[kerberos]
+        ver = "" if lib.match?(/^requests($|\[)/) || lib.start_with?("pyspnego")
+
+        final[lib] ||= {}
+        (final[lib][ver] ||= []) << mod
+      end
+    end
+
+    self
+  end
+
+  def output
+    result = final.flat_map do |lib, vers|
+      # consolidate multiple versioning rules
+      if vers.size > 1
+        max_key, *all_keys = vers.keys
+        all_keys.each do |alt|
+          higher, lower, conflict = version_compare(alt, max_key)
+          # There is a conflict when we have conflicting requirements. eg: >=2.0 and ==1.0
+          # We are displaying all comparisons/winners to verify the comparison algorithm works (skipping when merging a blank - no change of errors there)
+          $stderr.puts "#{lib}: #{higher} > #{lower} #{"CONFLICT" if conflict}" if lower != ""
+          vers[higher].concat(vers.delete(lower))
+          max_key = higher
+        end
+      end
+
+      vers.map do |(ver, modules)|
+        # if we pass in a previous requirements.txt, lets not mention it
+        # exception: display if it is only mentioned in a previous requirements.txt file
+        modules.delete("legacy") if modules.size > 1
+        "#{lib}#{ver} # #{modules.join(", ")}"
+      end
+    end.sort.join("\n")
+
+    puts result
+  end
+
+  private
+
+  def module_name_from_filename(fn)
+    if non_modules.include?(fn)
+      "legacy"
+    else
+      fn.gsub(%r{.*ansible_collections/}, "")
+        .gsub(%r{/requirements.*}, "")
+    end
+  end
+
+  def parse_line(line)
+    line.downcase!
+    # TODO: do we want to keep legacy comments? Only useful for our requirements.txt file
+    line.gsub!(/#.*/, "")
+    line.strip!
+    return if line.empty?
+
+    # Some libraries list "python" instead of "python_version"
+    # Dropping since just listing the python version isn't useful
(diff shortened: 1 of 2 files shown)


config/requirements.txt
```diff
@@ -1,45 +1,66 @@
...
 azure-mgmt-nspkg==2.0.0
-azure-mgmt-rdbms==1.4.1
-azure-mgmt-redis==5.0.0
-azure-mgmt-resource==2.1.0
-azure-mgmt-servicebus==0.5.3
-azure-mgmt-sql==0.10.0
-azure-mgmt-storage==3.1.0
-azure-mgmt-trafficmanager==0.50.0
-azure-mgmt-web==0.41.0
-azure-storage==0.35.1
```
kbrock, line 31: this was dropped (think replaced by azure-storage-blob

Don't see `azure-storage` listed in https://learn.microsoft.com/en-us/python/api/overview/azure/storage?view=azure-python

config/requirements.txt
```diff
@@ -1,51 +1,100 @@
...
-monotonic==1.4
-ncclient>=0.6.3
-netaddr==0.7.19
-openstacksdk==0.23.0
-ovirt-engine-sdk-python==4.2.4
-pexpect==4.6.0
-psutil==5.6.6 # Match the version installed directly into the venv
-pykerberos==1.2.1
+ciscoisesdk>=2.0.8 # cisco/ise
+cryptography>=36.0.0 # cisco/intersight, cisco/aci, microsoft/ad
```
Fryguy, line 55: This matches rpm as well which is `python3-cryptography-36.0.1-4.el9.x86_64`, so this is good.

config/requirements.txt
```diff
@@ -1,51 +1,100 @@
...
-ipaddress==1.0.23
-monotonic==1.4
-ncclient>=0.6.3
-netaddr==0.7.19
-openstacksdk==0.23.0
-ovirt-engine-sdk-python==4.2.4
-pexpect==4.6.0
-psutil==5.6.6 # Match the version installed directly into the venv
-pykerberos==1.2.1
+ciscoisesdk>=2.0.8 # cisco/ise
```
Fryguy, line 54: This makes me think we should drop support for cisco/ise and the ones we don't necessarily support since it's introducing new packages. cc @agrare

config/requirements.txt
```diff
@@ -1,51 +1,100 @@
...
+cryptography>=36.0.0 # cisco/intersight, cisco/aci, microsoft/ad
+deprecation # legacy
+dnacentersdk>=2.6.0 # cisco/dnac
+dnspython>=2.0.0 # microsoft/ad
+google-auth==1.6.2 # google/cloud
+google-cloud-storage # google/cloud
+grpcio # ansible/netcommon, cisco/iosxr
+ipaddress==1.0.23 # cisco/iosxr
```
Fryguy, line 62: I would love to drop ipaddress because it has had CVEs in the recent past, and we don't necessarily support cisco/iosxr

Fryguy:
This PR is fantastic - thanks for figuring this all out

Fryguy:
@kbrock Recapping our private convo with @agrare , let's drop all of the ciscos from this list except for cisco/intersight.

Fryguy:
> This downloads many versions of gssapi but keeps failing.

What is the error you're getting?  We might need to install a system package

kbrock:
@Fryguy It looks like we are missing `krb5-devel`, but we install that.

```
pip install 'pyspnego[kerberos]>=0.8.0'
Requirement already satisfied: pyspnego>=0.8.0 in /var/lib/manageiq/venv/lib/python3.9/site-packages (from pyspnego[kerberos]>=0.8.0) (0.11.1)
Requirement already satisfied: cryptography in /var/lib/manageiq/venv/lib64/python3.9/site-packages (from pyspnego>=0.8.0->pyspnego[kerberos]>=0.8.0) (43.0.1)
Collecting gssapi>=1.6.0 (from pyspnego[kerberos]>=0.8.0)
  Using cached gssapi-1.8.3.tar.gz (94 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      /bin/sh: line 1: krb5-config: command not found
```

bin/requirements.rb
```diff
@@ -0,0 +1,187 @@
...
+# And updates them for our supported packages
+#
+# USAGE:
+# 1. Get all module requirements (don't include documentation or testing ones)
+#
+#     ./requirements.rb ./requirements.txt /usr/lib/python3.9/site-packages/ansible_collections/ > new_requirements.txt
+#
+# 2. Resolve conflicts and determine if new one is correct
+#
+#     diff {,new-}requirements.txt
```
Fryguy, line 13: typo

```suggestion
#     diff {,new_}requirements.txt
```

bin/requirements.rb
```diff
@@ -0,0 +1,187 @@
...
+
+    self
+  end
+
+  def parse
+    filenames.each do |fn|
+      # the list of requirements-files can have items commented out - ignore those
+
+      mod = module_name_from_filename(fn)
+      IO.read(fn).split("\n").each do |line|
```
Fryguy, line 77: Doesnt really matter but:

```suggestion
      File.foreach(fn, chomp: true) do |line|
```

This may help if we encounter any windows line endings in the future.

bin/requirements.rb
```diff
@@ -0,0 +1,187 @@
...
+    line.downcase!
+    # TODO: do we want to keep legacy comments? Only useful for our requirements.txt file
+    line.gsub!(/#.*/, "")
+    line.strip!
+    return if line.empty?
+
+    # some libraries list "python" instead of "python_version"
+    line.gsub!(/^python /, "python_version")
+    # change version 5+ to >=5
+    line.gsub!(/\([0-9.]*\)[+]/, '>=\1')
```
Fryguy, line 146:
```suggestion
    line.gsub!(/\([0-9.]*\)\+/, '>=\1')
```

bin/requirements.rb
```diff
@@ -0,0 +1,187 @@
...
+    line.gsub!("= ", "=")
+    # ignore line where a module requires a minumum python version (assumption here)
+    return if line.match?(/^python_version/)
+    # ignore package requirements for older version of pythons (assumption here)
+    return if line.match?(/python_version ?[=<]/)
+
+    # split on first space (or =)
+    # version can have multiple spaces
+    lib, ver = line.match(/([^ >=]*) ?(.*)/).captures
```
Fryguy, line 155: I'm not sure I understand this line, but I think this can more simply be

```suggestion
    # split on first space (or =)
    # version can have multiple spaces
    lib, ver = line.split(/[ =>]/, 2)
```

bin/requirements.rb
```diff
@@ -0,0 +1,187 @@
...
+    # have already normalized for lowercase
+    # normalize library name with dash. all are the same
+    lib.gsub!(/[-_.]+/, "-")
+    ver ||= ""
+
+    [lib, ver]
+  end
+
+  # return highest, lowest for version comparison
+  def version_compare(a, b)
```
Fryguy, line 165: Can we use Gem::Version object here to do the heavy lifting or are the requirements not quite the same?  e.g. something like:

```suggestion
  def version_compare(a, b)
    [Gem::Version.new(a), Gem::Version.new(b)].sort.reverse
  end
```
```

Comment (Fryguy) [reply bin/requirements.rb:165]: or perhaps Gem::Requirement...

```ruby
a = Gem::Requirement.new(">=4.5.0")
a.requirements
# => [[">=", Gem::Version.new("4.5.0")]]

a = Gem::Requirement.new([">=4.5.0", ">=1.2.3"])
a.requirements
# => [[">=", Gem::Version.new("4.5.0")], [">=", Gem::Version.new("1.2.3")]]
```

Fryguy:
I have a handful of Ruby comments but they can all be follow ups if needed.

bin/requirements.rb
kbrock: nice. Tested with a few and it works for most everything but this breaks down:

```
library==3.4
```

It looses the first `=` and turns into `library=3.4`

Maybe we want something like:
```
"library==22 ; <3".split(/([>= ]+)/, 2)
```

I'd then rewrite the comparison routine to keep the comparator separate, though then again, it does work the way it is.

bin/requirements.rb
kbrock: Python and ruby seem very close here - https://pip.pypa.io/en/stable/reference/requirements-file-format/

I do like the idea of requirements instead of doing this ourselves.
But in the end of the day, there are only 2 instances where it is not a simple "merge no dependency with a dependency"

And I'm not sure that we want to invent a full requirements.txt parser in ruby to handle the `,` and `;` and all the wonderful `python_version` stuff.

kbrock:
update:
- Added venv information for setup
- Fixed new_requirements name typo
- Fixed add_target to better support directories and missing files
- Better "legacy" requirements detection
- Using `IO.foreach` (`File.foreach` didn't work)
- Capitalization for comments
- Smaller `python_version` line detection
- `;python_version` detection moved to `ver` processing (some day would be nice to improve)
- Fixed version 5+ regex
- pulled out `split_lib_ver`. Wasn't able to fix regexp - punting for now
-

kbrock:
ok, so with our list of requirements, we only have 1 or 2 version comparisons to make.

I think having a very rudimentary system and displaying the work so a person can double check the work makes the most sense.

Fryguy (approved):

Fryguy:
@kbrock This closed a lot of Mend issues 🎉 , but seems to have opened a few more. See https://github.com/ManageIQ/manageiq-rpm_build/issues?q=is%3Aissue+sort%3Aupdated-desc+updated%3A2024-09-18T02%3A50%3A00-04%3A00..2024-09-19+author%3Aapp%2Fmend-bolt-for-github

PR #495, opened 2024-09-06
Update ansible requirements.txt for vmware aws

kbrock:
CP4AIOPS-5619

When running ansible playbook against vmware or aws, it sometimes fails.
One issue: required python packages are not up to date.

The following warnings are displayed

```
[WARNING]: botocore < 1.21.0 is not supported or tested.  Some features may not
work.
[WARNING]: boto3 < 1.18.0 is not supported or tested.  Some features may not
work.

[WARNING]: Skipping plugin (/usr/lib/python3.9/site-
packages/ansible/plugins/connection/winrm.py), cannot load: deprecated() got an
unexpected keyword argument 'name'

The error was: AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'

"The installed version of pyvmomi lacks JSON output support; need pyvmomi>6.7.1"
```

Solution
---

### Requirement lines were updated.

- `/usr/lib/python3.9/site-packages/ansible_collections/amazon/aws/requirements.txt`
- `/usr/lib/python3.9/site-packages/ansible_collections/community/aws/requirements.txt`
- `/usr/lib/python3.9/site-packages/ansible_collections/community/vmware/requirements.txt`
- `/usr/lib/python3.9/site-packages/ansible_collections/vmware/vmware_rest/requirements.txt`

### Vmware requirements upgraded to support json

The `requirements.txt` listed `pyvmomi>=6.7`, but `6.7.1` is required for json.

### Requirements were loosened

There were warnings from pywinrm around deprecations. Loosened requirements to get newer version

- deprecation
- pywinrm

test1
-----

```yaml
---
- name: Run basic script to reproduce errors
  hosts: localhost
# amazon.aws - make sure this matches with the boto3 version
# put into roles/requirements.yml
  gather_facts: yes
  tasks:
    - name: ansible version
      debug: msg="ansible= {{ ansible_version.full }}"
    - name: python version
      debug: msg="python= {{ ansible_facts.python.version }}"
    - register: libraries
      shell: |
        ansible-galaxy collection list | grep aws
    - name: list installed aws libraries and versions
      debug:
        msg: " {{ libraries.stdout_lines }} "
    - name: view ec2 upgrade information
      command: ansible-galaxy collection list amazon.aws
    ###
    - name: load ec2_key module to display errors
      amazon.aws.ec2_key:
        name: key_pair
        state: present
        region: us-east-2
      register: result
    - name: keypair information
      debug:
        msg: "The EC2 key pair operation result: {{ result }}"
```

[...]

PR #503, opened 2024-09-06
Update azure keyvault to avoid parsing error

kbrock:
follow up to #495

- will upgrade azure version to a later one, but this alpha had a bad file in it and isn't working correctly
- this version of ncclient couldn't be resolved
- the git line should have read `git+https://...`
  But pip is currently not able to install this version


kbrock:
un-WIP: I was able to run all 3 test scripts locally.
Hope this means the env will work fine

Fryguy (approved):

PR #504, opened 2024-09-06
remove aiohttp

kbrock:
This was added to support all possible vmware clients But we don't use this client, none of our customers requested it and it looks to have security issues

@Fryguy your call if we want to go this route or want to resolve the issues

This resolves quite a few Mend issues (High and Medium) like https://github.com/ManageIQ/manageiq-rpm_build/issues/501 and https://github.com/ManageIQ/manageiq-rpm_build/issues/502

### DIFF
--- a/bin/requirements.rb
+++ b/bin/requirements.rb
@@ -36,7 +36,6 @@ class ParseRequirements
     openstack/cloud/requirements.txt
     ovirt/ovirt/requirements.txt
     theforeman/foreman/requirements.txt
-    vmware/vmware_rest/requirements.txt
   ]
   attr_reader :filenames, :non_modules, :final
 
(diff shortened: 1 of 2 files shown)


Fryguy:
I'm curious what @agrare thinks.  We didn't have it before, so it's not like it worked previously.

@agrare Background is there are 2 vmware clients for python.
- The `community/vmware` one requires pyvmomi and a git repo https://github.com/vmware/vsphere-automation-sdk-python.git
- The "official" one from vmware called `vmware/vmware-rest` requires aiohttp.

I'm pretty sure we've only every worked with community/vmware, but I wasn't sure if you had other experiences in the wild.

kbrock:
In that note. We've worked fine without the git package

kbrock:
I just added in https://github.com/ManageIQ/manageiq-rpm_build/pull/495
Think we want to drop this

PR #505, opened 2024-09-06
Add documentation on how to build a hotfix

agrare:
TODO:
- [x] Example of creating a patch with git format-patch and fixing up the directory prefixes to work with the builddir

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,28 @@ a series of patches.
 
 Create a patch file using `git format-patch` and copy to `rpm_spec/patches/`.  You will have to modify the path to be relative to the RPM BUILDDIR.
 
+For example if we wanted to create a hotfix to apply a commit in radjabov `1286aa34992340cad7a0778df01a845f1772e561` we would run:
+```sh
+git format-patch -1 1286aa34992340cad7a0778df01a845f1772e561
+```
+
+Which will create a patch file: `0001-Merge-pull-request-23123-from-agrare-fix_miq_request.patch`
+
+Make sure to remove any changes to spec files, as these are not packaged with the RPMs and will cause a conflict when
+the patch is applied.
+
+Next we have to change the file location to match the RPM BUILDDIR, for a patch in the core rpm this will be e.g. `manageiq-core-18.0-1`
+
+```
+diff --git a/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb b/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
+index 621f20b7e7..f62f454ac5 100644
+--- a/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
++++ b/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
+```
+
+This ensures that the hotfix patch will apply correctly when it is run.  Then move that patch
+file to `rpm_spec/patches/` in the rpm_build repo.
+
 `docker build --pull --tag $USER/rpm_build:radjabov-hotfix .`
 
 Create a `BUILD/hotfix` directory and copy the .src.rpm for the release that you want to
--- a/README.md
+++ b/README.md
@@ -127,6 +127,25 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
 docker rm ${CONTAINER}
 ```
 
+## Building a Hotfix
+
+Sometimes it is necessary to apply a patch without rebuilding the whole release in order to
+deliver a fix quickly.  This can be accomplished by building a hotfix rpm which will apply
+a series of patches.
+
+Create a patch file using `git format-patch` and copy to `rpm_spec/patches/`.  You will have to modify the path to be relative to the RPM BUILDDIR.
+
+`docker build --pull --tag $USER/rpm_build:radjabov-hotfix .`
+
+Create a `BUILD/hotfix` directory and copy the .src.rpm for the release that you want to
+build a hotfix against, for example `manageiq-release-18.0-1.el9.src.rpm`
+
+```sh
+docker run --rm -v `pwd`/OPTIONS:/root/OPTIONS -v `pwd`/BUILD:/root/BUILD $USER/rpm_build:radjabov-hotfix build_hotfix
+```
+
+The rpms for the hotfix will be in `BUILD/rpms`.
+
 ## Versioning
 
 Branch `morphy` == v13


README.md
```diff
@@ -127,6 +127,25 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
...
+
+Create a patch file using `git format-patch` and copy to `rpm_spec/patches/`.  You will have to modify the path to be relative to the RPM BUILDDIR.
+
+`docker build --pull --tag $USER/rpm_build:radjabov-hotfix .`
+
+Create a `BUILD/hotfix` directory and copy the .src.rpm for the release that you want to
+build a hotfix against, for example `manageiq-release-18.0-1.el9.src.rpm`
+
+```sh
+docker run --rm -v `pwd`/OPTIONS:/root/OPTIONS -v `pwd`/BUILD:/root/BUILD $USER/rpm_build:radjabov-hotfix build_hotfix
```
agrare, line 144: NOTE instead of a separate volume map for the hotfix dir it seemed easier to just map `./BUILD` so the artifacts and the srpm are in one place but if this doesn't work for some reason I can change it

Fryguy: I thought we already had this documented, but maybe not.  @bdunne Please review.

agrare: Yeah I didn't see any so this started as me just taking notes on what I had to do so we had something to reference in the future

README.md
```diff
@@ -127,6 +127,44 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
...
+Create a patch file using `git format-patch` and copy to `rpm_spec/patches/`.  You will have to modify the path to be relative to the RPM BUILDDIR.
+
+For example if we wanted to create a hotfix to apply a commit in radjabov `1286aa34992340cad7a0778df01a845f1772e561` we would run:
+```sh
+git format-patch 1286aa34992340cad7a0778df01a845f1772e561^..1286aa34992340cad7a0778df01a845f1772e561
+```
+
+Which will create a patch file: `0001-Merge-pull-request-23123-from-agrare-fix_miq_request.patch`
+
+Next we have to change the file location to match the RPM BUILDDIR, for a patch in the core rpm this will be e.g. `manageiq-core-18.0-1`
```
kbrock, line 145: I'm not totally sure what you did here.

Can you show us a diff of the `fix_miq_request.patch`?
Was it just a change to first line?

Yes, meta, a diff of a diff, but just trying to understand what was done here.
or maybe show it as a `sed` command?

README.md
kbrock: Having the docker command here is helpful

So much time goes between needing to build rpms that I often forget.

Re: "I thought we already had this documented"

On the mac, mapping the `BUILD` volume was horribly slow.

I remember putting together documentation on this a while back but people seemed to disagree on the best approach.

agrare: > On the mac, mapping the BUILD volume was horribly slow. Think I mapped BUILD/rpm to speed it up.
>
> I remember putting together documentation on this a while back but people seemed to disagree on the best approach.

Oh did you have hotfix documentation?  Or you mean documentation on how to build rpms on a mac?

README.md
bdunne: We also need to mention that you should remove any spec files from the patches.

README.md
```diff
@@ -127,6 +127,44 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
...
+
+Sometimes it is necessary to apply a patch without rebuilding the whole release in order to
+deliver a fix quickly.  This can be accomplished by building a hotfix rpm which will apply
+a series of patches.
+
+Create a patch file using `git format-patch` and copy to `rpm_spec/patches/`.  You will have to modify the path to be relative to the RPM BUILDDIR.
+
+For example if we wanted to create a hotfix to apply a commit in radjabov `1286aa34992340cad7a0778df01a845f1772e561` we would run:
+```sh
+git format-patch 1286aa34992340cad7a0778df01a845f1772e561^..1286aa34992340cad7a0778df01a845f1772e561
```
Fryguy, line 140: Simpler: 

```suggestion
git format-patch -1 1286aa34992340cad7a0778df01a845f1772e561
```

agrare: :+1: done

Fryguy:
@bdunne Please review.

README.md
Fryguy: @agrare Can you add the note about removing specs from the patch file?

agrare: :+1: done

README.md
```diff
@@ -127,6 +127,47 @@ docker cp ${CONTAINER}:/root/BUILD/rpms/x86_64/ ./rpms/
...
+diff --git a/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb b/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
+index 621f20b7e7..f62f454ac5 100644
+--- a/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
++++ b/manageiq-core-18.0-1/app/models/miq_provision_request_template.rb
+```
+
+This ensures that the hotfix patch will apply correctly when it is run.  Then move that patch
+file to `rpm_spec/patches/` in the rpm_build repo.
+
+`docker build --pull --tag $USER/rpm_build:radjabov-hotfix .`
```
jrafanie, line 160: FYI, I needed to supply the arch when building on my mac since we don't have aarch64 rpms in our manageiq rpm repository and my platform isn't the most common target arch: x86_64:

```
 podman build --pull --arch=x86_64 --tag $USER/rpm_build:radjabov-hotfix .
```

agrare: That's probably an issue for all of the docker examples in this README.  Maybe we split out a PR to update all of them to work on a mac?

jrafanie: oh for sure... I'm trying to validate it since I'm in this neck of the woods anyway.

agrare: Is podman the way to go on mac silicon macs now?  Maybe we can drop the docker commands here now too

jrafanie: Not sure.  That's what I'm using.  😉

README.md
jrafanie: My only addition is that I needed to manually clean the hotfix directory of everything  but the srpm.  It was listing the patch file multiple times in the rpm spec.  I'll open a PR.  We should just clean it before we copy over things.

jrafanie: sorry, one last thing... 

Note, I couldn't actually build a hotfix patch manageiq srpm.

Everything else seems to be working, although the [manageiq release value of 1](https://github.com/ManageIQ/manageiq-rpm_build/blob/master/rpm_spec/manageiq.spec.in#L30) in the spec doesn't line up with the expectation of a date format seen [here](https://github.com/ManageIQ/manageiq-rpm_build/blob/9ac5a5cc11153d304234086d98f5e18e3f3445bf/lib/manageiq/rpm_build/build_hotfix.rb#L45).  I had to use the product srpm to actually test building a hotfix.

README.md
jrafanie: FYI, the next command also worked with podman:

```
podman run --rm -v `pwd`/OPTIONS:/root/OPTIONS -v `pwd`/BUILD:/root/BUILD $USER/rpm_build:radjabov-hotfix build_hotfix
```

agrare: Yeah would be awesome if we could get one set of instructions that work for all, don't want to end up with another developer_setup with mac+ubuntu+centos yum+centos dnf for everything

Fryguy: I'm using podman desktop, but docker is still "common"

jrafanie (approved):
I'm good with this documentation.  It was very helpful for testing it.

podman or docker, it doesn't matter, it's easy enough to replace them vice versa.  We can decide on that later if needed.

Fryguy (approved):

PR #509, opened 2024-09-26
System rpm updates for paramiko and zipp

kbrock:
Updated 2 python packages and cleaned some dependencies up:

- Converted paramiko to an rpm dependency https://github.com/ManageIQ/manageiq-rpm_build/issues/430
- Convert python3-importlib-metadata (and therefore zipp) to an rpm dependency https://github.com/ManageIQ/manageiq-rpm_build/issues/476
- Dropped pexpect dependency. It is required by ansible-runner and not by a Collection.
- Cryptography is installed by rpm. Dropping explicit version.
- Dateutil is installed by rpm. Dropping explicit version.
- New build should resolve setuptools https://github.com/ManageIQ/manageiq-rpm_build/issues/477

Second commit fixes script to auto check rpms and no longer state package dependencies

### DIFF
--- a/bin/requirements.rb
+++ b/bin/requirements.rb
@@ -11,33 +11,23 @@
 #     source /var/lib/manageiq/venv/bin/activate
 #     chmod 755 requirements.rb
 #
-# 2. Determine all python packages provided by rpms (and compare with OS_PACKAGES)
-#
-#     TODO: may want to grep rpm contents with 'info$' to determine elgibility.
-#
-#     for pkg in $(rpm -qa | grep python3-) ; do echo "### $pkg" ; rpm -ql $pkg | awk  -F/ '/site-packages/ { print $6 }' | sort -u ; done
-#
-# 3. Get all module requirements (don't include documentation or testing ones)
+# 2. Get all module requirements
 #
 #     ./requirements.rb ./requirements.txt /usr/lib/python3.9/site-packages/ansible_collections/ > new_requirements.txt
 #
-# 4. Resolve conflicts and determine if new one is correct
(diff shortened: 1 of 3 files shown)


bin/requirements.rb
```diff
@@ -56,6 +46,17 @@ class ParseRequirements
   ]
   attr_reader :filenames, :non_modules, :final, :verbose
 
+  # These packages are installed via rpm
+  def os_packages
+    # Leaving his as pure bash so we can run from the command line to fix issues.
```
Fryguy, line 51:
```suggestion
    # Leaving this as pure bash so we can run from the command line to fix issues.
```

Fryguy (approved):

bin/requirements.rb
kbrock: fixed in #510

bin/requirements.rb
```diff
@@ -101,18 +102,16 @@ def add_dir(dirname)
...
         next unless lib
 
-        # skip git libraries. git>= line from vsphere gave us problems
+        # Skip git libraries. The 'git>=.*' line from vsphere gave us problems.
         next if lib.start_with?("git")
 
-        # system packages are versioned by rpms
-        ver = "" if lib.match?(/^(#{OS_PACKAGES.join("|")})($|\[)/)
+        # Defer to version requirements provided by rpm system packages.
+        ver = "" if lib.match?(/^(#{os_package_regex.to_s})($|\[)/)
```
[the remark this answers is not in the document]
kbrock: fixed in #510

PR #519, opened 2024-10-31
Update ansible azure storage blob

kbrock:
Update requirements to better solve these version issues in the future:
 
- rename bin/requirements to bin/parse_requirements
- Add tests for parse_requirements.rb (and refactored to make it easier to test)
- override `==` in requirements.txt to be `>=`


kbrock:
update:
- held off updating minimum versin for azure-storage-blob to 12.24.0 CVE-2022-30187 [ref](https://nvd.nist.gov/vuln/detail/CVE-2022-30187#range-12624282)

PR #524, opened 2024-11-22
Move paramiko into Dockerfile.

kbrock:
Before
======

Installs python paramiko package via rpm
Also installs paramiko python pip package.

After
=====

Ensure paramiko is available in the environment that builds the virtual env. virtual env building process will detect the package is available and and not try and install the paramiko pip package. So we end up with only 1 paramiko package installed. This will ensure we do not install the paramiko python pip package into the venv.

History
=======

This issue was introduced in
- https://github.com/ManageIQ/manageiq-rpm_build/pull/509
- 35401105fb1c8535


Dockerfile
```diff
@@ -57,6 +57,7 @@ RUN ARCH=$(uname -m) && \
       python3-pip \
       python3-virtualenv \
       rpm-build && \
+      python3-paramiko && \
```
Fryguy, line 60: This a) needs to go up alphabetically and then b) should not have `&&`

Fryguy (changes requested):

kbrock:
update:
- fixed location of `python3-paramiko` (and dropped the `&&`)

Dockerfile
```diff
@@ -55,6 +55,7 @@ RUN ARCH=$(uname -m) && \
       openssl-devel \
       python3-devel \
       python3-pip \
+      python3-paramiko \
```
Fryguy, line 58: Still not alphabetical :)

kbrock:
update:
- moved paramiko up in Dockerfile

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit fbbcdcc599ca86c856a3e6e9fcfe0b1ae55ebfb7.

```text
commit fbbcdcc599ca86c856a3e6e9fcfe0b1ae55ebfb7
Date:   Tue Nov 26 14:36:06 2024 -0500

    Merge pull request #524 from kbrock/paramiko_lib
    
    Move paramiko into Dockerfile.
    
    (cherry picked from commit b30f5e5a4a6cdf65e8405442e18be0d54f1bee77)
```

PR #537, opened 2025-01-02
Clean hotfix directory before patching the srpm contents

jrafanie:
For those of us who don't patch the srpm contents correctly the first or
even second time, we should do something like make clean before we extract
the srpm, patch things, and attempt to rpmbuild.

### DIFF
--- a/lib/manageiq/rpm_build/build_hotfix.rb
+++ b/lib/manageiq/rpm_build/build_hotfix.rb
@@ -14,6 +14,7 @@ module ManageIQ
         where_am_i
 
         Dir.chdir(HOTFIX_DIR) do
+          clean_hotfix_directory
           unpack_srpm
           update_spec
           copy_patches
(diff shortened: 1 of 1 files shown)


lib/manageiq/rpm_build/build_hotfix.rb
```diff
@@ -64,6 +65,11 @@ def update_spec
         File.write(rpm_spec, spec_text)
       end
 
+      def clean_hotfix_directory
+        files = Dir.glob("*").reject {|path| path.end_with?("src.rpm") }
```
agrare, line 69: Minor
```suggestion
        files = Dir.glob("*").reject { |path| path.end_with?("src.rpm") }
```

agrare: If we're not using the glob part of Dir.glob would Dir.children(".") work?

Fryguy: Event better:

```suggestion
        files = Dir.glob("*[!.src.rpm]")
```

jrafanie: haha, I had glob negation at first but thought this looked more obvious considering performance isn't going to matter on a handful of file.  I can change it if you also liked it 😉

jrafanie: haha, ship it... just verifying it

jrafanie: ok, testing looks fine

Fryguy (approved):

PR #556, opened 2025-03-06
[morphy] Lock down to 8 jobs for now

bdunne:
PPC has 64 cores, maybe that's causing the failures there?

### DIFF
--- a/lib/manageiq/rpm_build/generate_gemset.rb
+++ b/lib/manageiq/rpm_build/generate_gemset.rb
@@ -146,7 +146,7 @@ module ManageIQ
       def cpus
         c = `nproc --all` rescue nil
         c = c.to_i
-        c == 0 ? 3 : c
+        c.to_i.clamp(3, 8)
       end
 
       def cleanse_gemset


lib/manageiq/rpm_build/generate_gemset.rb
```diff
@@ -75,7 +75,7 @@ def populate_gem_home(build_type)
             shell_cmd("bundle lock --update --conservative --patch") if build_type == "nightly"
           end
 
-          shell_cmd("bundle install --jobs #{cpus} --retry 3")
+          shell_cmd("bundle install --jobs 8 --retry 10")
```
Fryguy, line 78: wonder if we should clamp it down in the cpus method directly and leave this line.

```diff
        def cpus
          c = `nproc --all` rescue nil
-         c = c.to_i
-         c == 0 ? 3 : c
+         c.to_i.clamp(3, 8)
        end
```

bdunne: Done

PR #562, opened 2025-05-05
Loosen psych dependency

bdunne:
Renovate is trying to update to <5.2.5 in #561 since 5.2.4 was just released. I don't see any reason for us to be locked down a z-stream, it was last changed from <4 to <5.2.4 in #528


Fryguy (approved):

Fryguy:
Backported to spassky via merge of master into spassky

PR #573, opened 2025-06-13
Upgrade ansible/ansible-core to latest and python to 3.12

Fryguy:
I've executed these tests in https://github.com/ManageIQ/manageiq/pull/23482 on containers and appliances before and after the RPM upgrade both with and without sourcing the manageiq-ansible-venv.

I've also figured out why ansible and ansible-core are not being removed on upgrade (we need `dnf autoremove` to remove)

Depends on
- [x] https://github.com/ManageIQ/manageiq/pull/23482

Follow ups
- Double check the content of the rpm and that there aren't duplicates with any python-3.12 system packages.


Fryguy:
>  Figure out why ansible and ansible-core are not being removed on upgrade

`dnf autoremove` will remove these, so we might want that as part of the upgrader script.

Fryguy:
@bdunne This is ready to go.

Issue #599, opened 2025-11-21
rpm build failing due to lack of disk space


jrafanie:
<img width="1276" height="845" alt="Image" src="https://github.com/user-attachments/assets/f1e05458-253b-47d6-b91a-b823a4ae09b5" />

It looks like it's downloading a bunch of rpms and running out of disk space.  Perhaps we had a purger that ran or we just went over threshold and now need a purger so we don't have so many rpms to download.

```
Downloading required RPMs...
Fetching uncached RPM: kafka-3.7.0-1.el9.src.rpm
Fetching uncached RPM: kafka-3.7.0-1.el9.x86_64.rpm
Fetching uncached RPM: manageiq-release-20.0-1.el9.noarch.rpm
Fetching uncached RPM: manageiq-release-20.0-1.el9.src.rpm
Fetching uncached RPM: python-bambou-3.1.1-1.el9.src.rpm
Fetching uncached RPM: python3-bambou-3.1.1-1.el9.noarch.rpm
Fetching uncached RPM: python-pylxca-2.1.1-2.el9.src.rpm
Fetching uncached RPM: python3-pylxca-2.1.1-2.el9.noarch.rpm
Fetching uncached RPM: python-unittest2-1.1.0-16.el9.src.rpm
Fetching uncached RPM: python-unittest2-1.1.0-24.el9.src.rpm
Fetching uncached RPM: python3-unittest2-1.1.0-16.el9.noarch.rpm
Fetching uncached RPM: python3-unittest2-1.1.0-24.el9.noarch.rpm
Fetching uncached RPM: python-vspk-5.3.2-1.el9.src.rpm
Fetching uncached RPM: python3-vspk-5.3.2-1.el9.noarch.rpm
Fetching uncached RPM: repmgr13-5.2.1-3.el9.src.rpm
Fetching uncached RPM: repmgr13-5.2.1-3.el9.x86_64.rpm
Fetching uncached RPM: repmgr13-debuginfo-5.2.1-3.el9.x86_64.rpm
Fetching uncached RPM: repmgr13-debugsource-5.2.1-3.el9.x86_64.rpm
Fetching uncached RPM: manageiq-20.0.0-20251109003053.el9.src.rpm
Fetching uncached RPM: manageiq-20.0.0-20251110002932.el9.src.rpm
Fetching uncached RPM: manageiq-20.0.0-20251111002523.el9.src.rpm
[...]

PR #601, opened 2025-11-24
Split the RPM build and the RPM repo metadata update

bdunne:
- Ensure we are deleting the build container after it runs and uploads the RPMs
- The RPM repo update does not require the container and can be run directly with the system ruby.

If I run the update_rpm_repo.rb script locally, the downloaded content is just over 5GB, so that alone shouldn't run out of disk space.  We're probably running out of space because we have all of the other files generated by the RPM build itself inside the container.

Fixes: https://github.com/ManageIQ/manageiq-rpm_build/issues/599

### DIFF
--- a/.github/workflows/build_rpms.yaml
+++ b/.github/workflows/build_rpms.yaml
@@ -23,6 +23,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
+    - name: Set up Ruby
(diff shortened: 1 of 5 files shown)


jrafanie:
> WIP:
> 
>     * Verify that RPMs will still be pushed
> 
>     * Also run the purger

Is this ready to go @bdunne?  It looks good so far.

PR #627, opened 2026-03-11
Fix restorecon clearing valid selinux labels

agrare:
If there are files in `/var/lib/manageiq/containers/storage` when running the rpm install then the selinux context of the subdirectories gets wiped out.

Since we're creating this directory it should be empty and there should be no need to run restorecon.  Once the directory is created all files will have proper contexts anyway and relabeling also not necessary.

This fixes any existing files in `/var/lib/manageiq/containers/storage` having the `container_var_lib_t` context after running the rpm install.


Fryguy:
Backported to `tal` in commit 5781fc2026ff82cdb691d597a3c04c38e598876e.

```text
commit 5781fc2026ff82cdb691d597a3c04c38e598876e
Date:   Wed Mar 11 16:16:20 2026 -0400

    Merge pull request #627 from agrare/drop_restorecon_from_manageiq_core
    
    Fix restorecon clearing valid selinux labels
    
    (cherry picked from commit f4d1a585a24e89b4a589c7f7cbd3a7e885779f68)
```

PR #629, opened 2026-03-12
Set SELinux context non-recursively for manageiq containers/storage

agrare:
Ensure that the manageiq containers/storage directory get the proper selinux context after it is created.  Don't run restorecon recursively which was an issue previously, only on these two folders which get created if they are missing.

I realized a flaw in my testing of the first case here https://github.com/ManageIQ/manageiq-rpm_build/pull/627#issuecomment-4040458898 where I deleted the directory so it had to be created, but I didn't account for the selinux contexts already existing from the existing installation.

To test this I deleted the containers/storage dir and also all of the selinux contexts:
```
rm -rf /var/lib/manageiq/containers
for f in `semanage fcontext --list | grep 'var/lib/manageiq' | awk '{print $1}'`; do echo $f; semanage fcontext --delete $f; done
```

This would occur if someone upgrades across multiple versions from when this directory didn't exist.

### DIFF
--- a/rpm_spec/subpackages/manageiq-core
+++ b/rpm_spec/subpackages/manageiq-core
@@ -72,6 +72,9 @@ semanage fcontext --add --type container_ro_file_t '%{manageiq_var_lib_root}/con
 semanage fcontext --add --type container_file_t    '%{manageiq_var_lib_root}/containers/storage/volumes/[^/]*/.*'
 semanage fcontext --add --type container_var_lib_t '%{manageiq_var_lib_root}/containers(/.*)?'
 
+restorecon %{manageiq_var_lib_root}/containers/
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

Fryguy:
Backported to `tal` in commit ca414bf9094106245187b09420896db906b8be10.

```text
commit ca414bf9094106245187b09420896db906b8be10
Date:   Thu Mar 12 13:46:58 2026 -0400

    Merge pull request #629 from agrare/restorecon_containers_storage
    
    Set SELinux context non-recursively for manageiq containers/storage
    
    (cherry picked from commit d44ae0f6c9e82e844e27f887c347376711372734)
```

PR #628, opened 2026-03-11
Repmgr16 1.5.5 build

jrafanie:
* Update various references with 13 to 16
* Update to latest repmgr, 1.5.5
* Enhance debugging of builds by capturing the mock logging on error and printing it in the GHA output.
* Add missing buildRequires for repgmr 1.5.5 on postgresql 16
* Fix some mock deprecations/warnings/failures

TODO:

- [x] Build the rpms and test them on a postgresql 16 manageiq appliance
- [x] Fix any bugs in our HA/region replication code that had hardcoded pathing for repmgr or similar.

PR #630, opened 2026-03-12
Upgrade to EL10

bdunne:
Part of: https://github.com/ManageIQ/manageiq/issues/23737


rpm_spec/subpackages/manageiq-gemset
```diff
@@ -31,9 +31,6 @@ Requires: qpid-proton-c >= 0.37.0
 # For Appliance Console
 Requires: libsodium-devel
 
-# For scanning
-Requires: glusterfs-fuse
```
bdunne: Gluster is dead and not available in CentOS Stream 10

bdunne:
I squashed and rebased my commits to clean things up and also pulled in the changes from Joe's PR.  I'm going to do another full build tomorrow to verify that all of the packages are good to go.

rpm_spec/subpackages/manageiq-gemset
Fryguy: @agrare What will this affect provider-wise - it says for scanning, but is that for smartstate or something else like openscap scanning?

agrare: Ovirt/RHV SSA:
* https://github.com/ManageIQ/manageiq-smartstate/blob/master/lib/MiqVm/MiqRhevmVm.rb#L159C11-L159C30
* https://github.com/ManageIQ/manageiq-gems-pending/tree/master/lib/gems/pending/util/mount/miq_glusterfs_session.rb

bdunne:
Ok, I think this is ready for review.

packages/kafka/kafka.spec
```diff
@@ -1,84 +1,97 @@
 Name:             kafka
 Summary:          Apache Kafka is an open-source stream-processing software platform
-Version:          3.7.0
+Version:          3.9.1
 Release:          1%{?dist}
 License:          Apache (v2)
 Group:            Applications
 URL:              https://kafka.apache.org
-BuildRequires:    java-11-openjdk-devel
+BuildRequires:    git
```
bdunne, line 8: For the git patch

rpm_spec/subpackages/manageiq-gemset
bdunne: The most recent builds of gluster are for EL9 and years old at this point

bdunne:
Ok @jrafanie @Fryguy I think this is ready to go

rpm_spec/subpackages/manageiq-gemset
agrare: I assume we'll want to drop glusterfs support from RHV SSA then right?

packages/kafka/kafka.spec
Fryguy: Minor, but I thought patch was it's own CLI and didn't need git. On that note where does the patching happen (is that the `autosetup -S git`?)

rpm_spec/subpackages/manageiq-gemset
Fryguy: yeah we'll have to drop gluster support for SSA.

Fryguy:
Not sure how we can merge this before dropping the gluster support for RHV SSA (@agrare Unless you think we can merge it anyway?)

Fryguy:
@bdunne This PR also moves the kafka stuff from opt to etc to support bootc. But how will that affect upgrades? I'm concerned we introduce that and then upgrades are broken (although I admit this is a full appliances replacement, so maybe that's fine).

agrare:
> Not sure how we can merge this before dropping the gluster support for RHV SSA (@agrare Unless you think we can merge it anyway?)

Agreed I think we need to drop that first then we could merge this (or at least make sure we drop it before we release).

I'm a little surprised this package isn't in EL10, Red Hat Gluster might be gone but glusterfs the project isn't, and oVirt still supports gluster storage.

bdunne:
> @bdunne This PR also moves the kafka stuff from opt to etc to support bootc. But how will that affect upgrades? I'm concerned we introduce that and then upgrades are broken (although I admit this is a full appliances replacement, so maybe that's fine).

Yeah, this would only be on EL10 appliances where the upgrade path is an appliance replacement, so I don't think that's an issue.

rpm_spec/subpackages/manageiq-gemset
agrare: https://github.com/ManageIQ/manageiq-smartstate/pull/241

Dockerfile
```diff
@@ -7,23 +7,15 @@ ENV TERM=xterm \
...
-      http://mirror.stream.centos.org/9-stream/BaseOS/${ARCH}/os/Packages/centos-stream-release-9.0-34.el9.noarch.rpm \
-      http://mirror.stream.centos.org/9-stream/BaseOS/${ARCH}/os/Packages/centos-stream-repos-9.0-34.el9.noarch.rpm \
-      http://mirror.stream.centos.org/9-stream/BaseOS/${ARCH}/os/Packages/centos-gpg-keys-9.0-34.el9.noarch.rpm && \
-    dnf -y install \
-      https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm \
-      https://rpm.manageiq.org/release/21-uhlmann/el9/noarch/manageiq-release-21.0-1.el9.noarch.rpm && \
-    dnf -y --disablerepo=ubi-9-baseos-rpms swap openssl-fips-provider openssl-libs && \
-    dnf -y update && \
-    dnf -y module enable ruby:3.3 && \
-    dnf -y module enable nodejs:22 && \
```
jrafanie, line 24: Do we need to reset/enable pg:16?

This is what I tested with a few weeks ago in https://github.com/ManageIQ/manageiq-rpm_build/pull/628

```
    dnf -y module reset postgresql && \
    dnf -y module enable postgresql:16 && \
```

Fryguy:
I can merge this before we do the SSA drop since it will all land in the release.

Dockerfile
Fryguy: Spoke to Brandon and in short no.

bdunne: No. 16 is the default long term supported version on EL10, but if you prefer, we could do [the same thing that the container-postgresql does.](https://github.com/ManageIQ/container-postgresql/blob/163a34702c48dbf254a52b287a0c113243d15cec/Dockerfile#L76)

PR #652, opened 2026-04-07
[WIP] Upgrade kafka to 3.9.1 and changes for EL10

bdunne:
- Use Systemd sysusers rather than custom logic for creating the kafka user and group and set a static UID to avoid future upgrade issues.
- Move kafka and zookeeper configuration to `/etc/kafka` (`/opt` is read-only in bootc)
- Update to java 21.  Gradle supports Java 23, but the next packaged version on EL10 is Java 25.
- Upgrade Kafka to the latest 3.x version (4.x switches to Kraft.  We may want to do that soon, but I didn't want to do it in this PR.)
- Move config changes from `sed` in the spec file to a proper patch.


bdunne:
Merged as part of #630

PR #656, opened 2026-04-13
[WIP] Use versioned pip when installing ansible for seeding

Fryguy:
@bdunne Please review.

WIP as I need to test locally.

### DIFF
--- a/lib/manageiq/rpm_build/generate_core.rb
+++ b/lib/manageiq/rpm_build/generate_core.rb
@@ -65,7 +65,7 @@ module ManageIQ
       end
 
       def seed_ansible_runner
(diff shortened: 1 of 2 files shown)


lib/manageiq/rpm_build/generate_core.rb
```diff
@@ -65,7 +65,7 @@ def build_service_ui
       end
 
       def seed_ansible_runner
-        shell_cmd("pip install ansible") # This should happen after the ansible-venv is fully built
+        shell_cmd("#{GenerateAnsibleVenv.pip_versioned} install ansible") # This should happen after the ansible-venv is fully built
```
Fryguy, line 68: I also don't think we need the entirety of ansible and might be able to just do ansible-core - That could save some build time to not install all that.

PR #657, opened 2026-04-14
Drop nuage specific packages

agrare:
Related:
* https://github.com/ManageIQ/manageiq/issues/23759


rpm_spec/subpackages/manageiq-gemset
```diff
@@ -22,10 +22,9 @@ Requires: ipmitool
 # For Lenovo
 Requires: python3-pylxca
 
-# For Nuage
+# For Qpid Proton
 Requires: cyrus-sasl
 Requires: cyrus-sasl-plain
```
agrare, line 27: WIP pending investigating if these two are for qpid-proton-c or something nuage specific

I know we use SASL for e.g. rdkafka so this might be used elsewhere

agrare: Confirmed cyrus-sasl is a requirement for qpid-proton-c (not sure if we need to declare them separately then but we can look into that separately I'd rather not break anything)
https://rpmfind.net/linux/RPM/fedora/devel/rawhide/x86_64/q/qpid-proton-c-0.40.0-13.fc44.x86_64.html

agrare:
Taking out of WIP

/cc @bdunne @Fryguy

packages/python-bambou/build
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-
-set -e
-
-curl --fail -o bambou-3.1.1.tar.gz https://rpm.manageiq.org/sources_cache/python-bambou/bambou-3.1.1.tar.gz
```
Fryguy, line 5: @bdunne Do we keep the sources cache too in case we need it for an older branch, or should we drop that too?

packages/python-vspk/build
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-
-set -e
-
-curl --fail -o vspk-5.3.2.tar.gz https://rpm.manageiq.org/sources_cache/python-vspk/vspk-5.3.2.tar.gz
```
Fryguy, line 5: Ditto

rpm_spec/subpackages/manageiq-gemset
Fryguy: It looks like it requires on cyrus-sasl-lib, but I'm not sure why we have `cyrus-sasl`. `cyrus-sasl-plain` makes sense if we have plain formats for SASL that we need

Fryguy: FYI: https://github.com/ManageIQ/manageiq-appliance-build/pull/228 was when it was added.

> The cyrus-sasl packages are required to support the authentication mode Nuage network provider uses in their validated deployments.

agrare: Ah okay, I tracked the commit from Satoe that added this here but it didn't say where it came from

Fryguy: Did you find that openstack needed these extra cyrus-sasl packages? I think it's internal dependency on cyrus-sasl-lib is probably sufficient and I don't think we need these extra packages.

(Though I agree we can remove them separately - I just wasn't sure if you saw a hard need for them)

agrare: I don't know, I just saw the cyrus-sasl-libs dependency in qpid-proton-c and thought it safer to leave anything related to it here.

agrare: I also don't have an openstack with STF to be able to test it.

Fryguy (approved):
LGTM - just a couple questions on the sources for @bdunne

packages/python-bambou/build
bdunne: Yeah, I'll keep that, it's small.  This PR looks good.

PR #659, opened 2026-04-17
Update node 22

GilbertCherrie:
Part of: https://github.com/ManageIQ/manageiq/issues/23810
Update Node version to Node 22.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ RUN ARCH=$(uname -m) && \
       libxslt-devel \
       libyaml-devel \
       nodejs \
+      npm \
       openssl-devel \
       platform-python-devel \
       postgresql-server \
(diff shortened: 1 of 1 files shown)


GilbertCherrie:
@miq-bot assign @Fryguy 
@miq-bot add-label dependencies

Fryguy:
@bdunne This one is weird - Does nodejs 22 not come with npm? According to https://www.rpmfind.net/linux/RPM/almalinux/9.7/appstream/x86_64/nodejs-22.22.2-1.module_el9.7.0+227+12323393.x86_64.html, it should be in there.

Fryguy:
Ok, I checked locally, and npm just isn't installed - I think they split it out to a separate package. @GilbertCherrie Can you add `npm` to the list of packages we install with dnf install?

When i added that locally I was able to get the build container to build.

GilbertCherrie:
@Fryguy yeah just updated it

PR #660, opened 2026-04-21
[WIP] Leverage rpm createrepo --cachedir for less storage usage on builds

Fryguy:

### DIFF
--- a/bin/initialize_rpm_cache.rb
+++ b/bin/initialize_rpm_cache.rb
@@ -1,0 +1,92 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+
+require 'bundler/setup'
+require 'manageiq-rpm_build'
+
+module ManageIQ
+  module RPMBuild
+    class RpmCacheInitializer
+      include Helper
+      include S3Common
+
+      def initialize
+        @cache_dir = ROOT_DIR.join("rpm_cache")
+      end
+
+      def run
+        require 'tempfile'
+        Dir.mktmpdir do |tmpdir|
+          work_dir = Pathname.new(tmpdir)
+
+          # Build directory structure /release/{master,jansa}/el9/{noarch,src,x86_64}
+          directories = OPTIONS.rpm_repository.content.flat_map do |branch, values|
+            values[:targets].product(OPTIONS.rpm_repository.arches).map do |target, arch|
+              work_dir.join("release", branch.to_s, target, arch)
+            end
+          end
+          require 'fileutils'
+          directories.each { |dir| FileUtils.mkdir_p(dir) }
+
+          # Fetch ALL configured RPMs (one-time download)
+          puts "Downloading ALL RPMs for initial cache generation..."
+          OPTIONS.rpm_repository.content.each do |branch, values|
+            values[:rpms]&.each do |rpm, version_regex|
+              client.list_objects(:bucket => OPTIONS.rpm_repository.s3_api.bucket, :prefix => File.join("builds", rpm.to_s, "/")).flat_map(&:contents).each do |object|
+                file = object.key
+                name = File.basename(file)
+                *, target, arch, _rpm = name.split(".")
+                next unless values[:targets].include?(target)
+                next unless OPTIONS.rpm_repository.arches.include?(arch)
+                next unless version_regex =~ name
+
+                cached_rpm = @cache_dir.join(name)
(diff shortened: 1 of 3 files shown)


bin/initialize_rpm_cache.rb
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path("../lib", __dir__)
+
+require 'bundler/setup'
+require 'manageiq-rpm_build'
+
+module ManageIQ
+  module RPMBuild
+    class RpmCacheInitializer
```
Fryguy, line 10: This is a one-time script for building the first cache dir - we wouldn't commit this we'd just run it locally.

docs/rpm_repo_caching.md
```diff
@@ -0,0 +1,118 @@
...
+├── builds/                    # Source RPMs
+├── release/                   # Published repository
+│   ├── 21-uhlmann/
+│   │   └── el9/
+│   │       ├── noarch/
+│   │       ├── src/
+│   │       └── x86_64/
+│   │           ├── *.rpm
+│   │           └── repodata/
+└── createrepo_cache/          # RPM metadata cache
```
Fryguy, line 80: I wasn't sure if we wanted this top-level or if we wanted to this within each release repo - I'm not sure what the common pattern is for other RPM repositories.

docs/rpm_repo_caching.md
```diff
@@ -0,0 +1,118 @@
...
+
+- [`bin/initialize_rpm_cache.rb`](../bin/initialize_rpm_cache.rb) - One-time initialization script
+- [`lib/manageiq/rpm_build/rpm_repo.rb`](../lib/manageiq/rpm_build/rpm_repo.rb) - Updated repository update logic
+
+## Troubleshooting
+
+### Cache Corruption
+If the cache becomes corrupted, you can regenerate it by running:
+```bash
+./bin/initialize_rpm_cache.rb
```
Fryguy, line 96: This might be the one reason we want to keep the script committed.

I also wasn't sure what we do when we delete RPMs - the cache hash would stick around but maybe that's ok?  Do we still delete RPMs on a time-based delete, or is it a script? (I can't remember)

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -31,50 +38,99 @@ def update
                 next unless values[:targets].include?(target)
                 next unless OPTIONS.rpm_repository.arches.include?(arch)
                 next unless version_regex =~ name
+
+                destination = work_dir.join("release", branch.to_s, target, arch, name)
```
bdunne, line 42: Not sure why this was moved

lib/manageiq/rpm_build/rpm_repo.rb
```diff
@@ -31,50 +38,99 @@ def update
...
+          # Upload updated repodata
+          Dir.glob(work_dir.join('release', '**', 'repodata', '*')).each do |file|
             next unless File.file?(file)
             destination_name = file.sub("#{work_dir.to_s}/", '')
-
             uploaded_files << destination_name if upload_file(file, destination_name)
           end
 
-          # Cleanup old stuff online
-          puts "Cleaning old files in bucket..."
```
bdunne, line 67: The cleanup was removed?

docs/rpm_repo_caching.md
```diff
@@ -0,0 +1,118 @@
+# RPM Repository Caching Implementation
```
bdunne, line 1: Wow, quite a 📖 (lots of duplication)

bin/initialize_rpm_cache.rb
bdunne: I'm fine with it being committed in case we need to regenerate it in the future for some reason.

docs/rpm_repo_caching.md
bdunne: I think global is okay for now.  We can adjust later if needed.

docs/rpm_repo_caching.md
bdunne: It's a script that I run occasionally.  I think we set the expiration time at upload and IBM COS ignores it.

docs/rpm_repo_caching.md
Fryguy: Yeah it's all AI haha. I probably won't commit this in the final thing

PR #673, opened 2026-05-19
Inline the container image build for update_ansible_reqs

Fryguy:
In trying to do #672 , I realized it doesn't buy us much, because inlining does the same thing. For now, let's inline and if we want to extract we can do that in a different PR.

### DIFF
--- a/.github/workflows/update_ansible_reqs.yaml
+++ b/.github/workflows/update_ansible_reqs.yaml
@@ -9,7 +9,7 @@ concurrency:
 permissions:
   contents: read
 jobs:
-  build_rpm_build_container:
+  update_ansible_reqs:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_ansible_reqs.yaml
```diff
@@ -20,12 +20,6 @@ jobs:
         echo "REGISTRY_PASSWORD=${{ secrets.DOCKER_REGISTRY_PASSWORD }}" >> $GITHUB_ENV
     - name: Build and push RPM build image
```
Fryguy, line 21: Note that we don't really need to push it either - I'm tempted to split the build and push steps to simplify this further, but not for this PR.

PR #686, opened 2026-07-02
Add CI workflow to validate config/requirements.txt changes

Fryguy:
Adds spec/test_ansible_reqs which builds a Python 3.12 venv at /var/lib/manageiq/venv, installs ansible-runner and requirements.txt, then runs the [ansible-runner] CLI tests from ManageIQ/manageiq's runner_execution_spec.bats (via a sparse checkout).

Adds .github/workflows/ci-ansible_reqs.yaml to trigger this only when config/requirements.txt is changed on PRs or pushes.

Depends on
- [x] https://github.com/ManageIQ/manageiq/pull/23910

Needed by https://github.com/ManageIQ/manageiq-rpm_build/pull/679

@bdunne Please review.

### DIFF
--- a/spec/test_ansible_reqs
+++ b/spec/test_ansible_reqs
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+set -e
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+
+PYTHON_VERSION="3.12"
+VENV_PATH="/var/lib/manageiq/venv"
+
+# Build the venv at the production path (runner_execution_spec.bats hard-codes this)
+echo -e "\033[1;34mCreating venv at $VENV_PATH...\033[0m"
+sudo mkdir -p /var/lib/manageiq
+sudo chown "$USER" /var/lib/manageiq
+python${PYTHON_VERSION} -m venv --system-site-packages --upgrade-deps $VENV_PATH
+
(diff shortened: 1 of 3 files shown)


config/requirements.txt
Fryguy, line 1: This should be removed before merge, but I need a change in the file so the workflow actually executes, since it's path based.

spec/test_ansible_reqs
```diff
@@ -0,0 +1,45 @@
...
+$VENV_PATH/bin/pip${PYTHON_VERSION} install --no-compile -r config/requirements.txt
+
+# Fetch test_ansible_runner_execution and the ansible specs from ManageIQ/manageiq.
+# We use a sparse checkout to avoid cloning the full repo.
+if [ ! -d "./spec/lib/ansible" ]; then
+  echo -e "\033[1;34mCloning ansible specs from ManageIQ/manageiq...\033[0m"
+  rm -rf /tmp/manageiq
+  # git clone --no-checkout --depth=1 --filter=tree:0 https://github.com/ManageIQ/manageiq /tmp/manageiq
+  # TODO REVERT ME
+  git clone --no-checkout --depth=1 --filter=tree:0 --branch=test_ansible_runner_filter https://github.com/Fryguy/manageiq /tmp/manageiq
```
Fryguy: Revert this after the core PR is merged.

Fryguy:
@bdunne This is ready to go.

